PR #2, opened 2016-08-25
feat(api): a minimal working swagger server

jackfrancis:
with README updates, and Makefile boilerplate


jackfrancis:
@migmartri @prydonius here is a minimal swagger-generated API, with plenty of boilerplate befitting a newly minted repo

rootfs/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:3.3
+
+RUN apk -U add ca-certificates && \
+    rm -Rf /var/cache/apk/*
+EXPOSE 8081
+ENV PORT 8081
+ENV HOST 0.0.0.0
+COPY usr/bin/monocular /usr/local/bin/monocular
```
migmartri: I am wondering if instead of assuming that the binary is in the host, make this dockerfile to build the binary from the source code.

migmartri:
Awesome start, I still need to wrap my head around what is own code vs generated code.

api/swagger-spec/swagger.yml
```diff
@@ -0,0 +1,20 @@
+swagger: "2.0"
```
migmartri: Are you using a swagger editor to create this file or you are writing the yaml files from scratch?

jackfrancis: This one was written manually.

PR #3, opened 2016-08-25
feat(healthz): basic healthz handler implementation

jackfrancis:
This includes an example handler implementation of our initial `GET /healthz` API endpoint. I intentionally did this as a separate PR to demonstrate some of the flow of swagger-generated dev.

### DIFF
--- a/pkg/swagger/restapi/configure_monocular.go
+++ b/pkg/swagger/restapi/configure_monocular.go
@@ -7,6 +7,7 @@ import (
 	httpkit "github.com/go-swagger/go-swagger/httpkit"
 	middleware "github.com/go-swagger/go-swagger/httpkit/middleware"
 
+	"github.com/helm/monocular/handlers"
 	"github.com/helm/monocular/pkg/swagger/restapi/operations"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/swagger/restapi/configure_monocular.go
```diff
@@ -25,7 +26,7 @@ func configureAPI(api *operations.MonocularAPI) http.Handler {
 	api.JSONProducer = httpkit.JSONProducer()
 
 	api.HealthzHandler = operations.HealthzHandlerFunc(func() middleware.Responder {
-		return middleware.NotImplemented("operation .Healthz has not yet been implemented")
+		return handlers.Healthz()
```
jackfrancis: `configure_monocular.go` (which swagger generates only if it's not already there) is where you glue in the actual implementation. This file is resilient against future swagger generation as the API evolves.

Although in fact some API configuration changes will not be enabled unless you let swagger "re-generate" the config, which will then include new boilerplate scaffolding. In those cases I've renamed the original config file, regenerated, and then manually interpolated the manual implementation code into the newly generated config.

PR #5, opened 2016-09-07
chore(dirs): reorganized root structure to accommodate ui codebase

jackfrancis:
API code, including all build and dependency management matter (`Makefile`, `glide.yaml`, among others), moved into `src/api`. All API building and testing will need to `cd` into that directory.


src/api/pkg/swagger/models/chart_file.go
```diff
@@ -0,0 +1,173 @@
+package models
```
prydonius: is this a new file?

jackfrancis: thanks!, that's cruft (this is swagger not deleting old generated stuff if you delete it from the spec)

jackfrancis: @prydonius just pushed a 2nd commit to delete that file

PR #6, opened 2016-09-08
feat(data): charts as generic API resources

jackfrancis:
including a few different things here:

- added a generic "resource" model via swagger that we'll use to represent all resource data responses from the API
- fully fleshed out a "ChartVersion" model that will represent a versioned chart as returned by a repo (replaced the prior skeleton "Chart" model)
- added some conveniences to help convert YAML to our data models
- added a mock "stable" repo (via a YAML file, and added conveniences which will be reusable when we eventually connect to actual HTTP repositories
- all getters (including mocks) are now general-purpose


src/api/data/helpers/helpers.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+	if latest == "" {
+		return ret, fmt.Errorf("unable to determine latest version")
+	}
+	return ret, nil
+}
+
+// newestSemVer returns the newest (largest) semver string
+func newestSemVer(v1 string, v2 string) (string, error) {
```
migmartri: What do you think about using a Semver specific package for version manipulation? I am not sure if the code below will work with pre-releases and build numbers.

In Helm we are using github.com/Masterminds/semver

src/api/swagger-spec/swagger.yml
```diff
@@ -38,47 +38,94 @@ paths:
...
+        type: string
+        minLength: 1
+      id:
+        type: string
+        minLength: 1
+      links:
+        type: object
+      attributes:
+        type: object
+  chartVersion:
```
migmartri: Is here where a new property key for the readme.md content will be added?

migmartri:
Added a couple of comments but nothing that should block the merge of this PR. Good job!! :)

src/api/data/helpers/helpers.go
jackfrancis: Yep, 100% agree. I'll add an issue for this.

src/api/swagger-spec/swagger.yml
jackfrancis: I think at this time where we put the readme.md is implementation-TBD. It could be a part of the above model, or it could be put as an `attribute` key/val into `resource` model as we prepare the HTTP response.

src/api/data/helpers/helpers.go
jackfrancis: https://github.com/helm/monocular/issues/7

PR #10, opened 2016-09-09
Monocular UI bootstrap

migmartri:
Initial scaffold for the Monocular client web application.

###  Chart page

selection_141

### Search result
selection_140

### Directory of charts
selection_139


jackfrancis:
👍

PR #12, opened 2016-09-16
feat(mocks): added lots more charts mock data

jackfrancis:
Added more mock charts data, so we have more to work with.

@migmartri @prydonius FYI when integrating with the UI: there is a `data.attributes.created` property present in the API JSON response to a charts resource that I didn't notice in the existing UI mock data. (Just in case that throws off some kind of pre-existing proto-validation code or something.)


migmartri:
Thanks for the heads up :) LGTM

PR #14, opened 2016-09-27
chore(swagger): upgraded to 0.6.0

jackfrancis:
This release includes several breaking changes, which is why it benefits us to do it now in the early prototype phase.

A few key changes that related to monocular:

- All route path handler funcs now require a params object, this is so we can inspect HTTP headers whether or not we're parsing URL params.
- All required model fields are now pointers (whether or not they functionally "require" having nil value representation).
- Several libraries have been replaced by `github.com/go-openapi` equivalents.

See the release page for full deets:

https://github.com/go-swagger/go-swagger/releases/tag/0.6.0

Glide and Makefile updated appropriately.


src/api/data/helpers/helpers.go
```diff
@@ -106,3 +106,13 @@ func newestSemVer(v1 string, v2 string) (string, error) {
 	}
 	return v1, nil
 }
+
+// Int64Point converts an int64 to an *int64
+func Int64Point(n int64) *int64 {
```
technosophos: Super pedantic, and I don't really care, but `Ptr` is the usual abbreviation in Go for a pointer. Again, not at all a showstopper. Feel free to ignore.

technosophos:
Ouch! The joys of a major upstream refactor.

PR #1, opened 2016-08-24
introductory and disclaimer matter

jackfrancis:

PR #20, opened 2016-10-04
WIP - UI / Backend integration

migmartri:
- [UI] Charts API backend fetching
- CORS support, enabled by config.yaml file and dependent on ENVIRONMENT env variable
- Docker multi-tier development environment (docker-compose up from root dir)


dev_env/api/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM quay.io/deis/go-dev:0.17.0
```
migmartri: Using the same go image you are using for `make build`

dev_env/api/rootfs/app-entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+dependencies_up_to_date() {
+  # It it up to date if the package file is older than
+  # the last time the container was initialized
+  [ ! $PACKAGE_FILE -nt $INIT_SEM ]
+}
+
+if [ "$1" == "go" -a "$2" == "run" ]; then
+	if ! dependencies_up_to_date; then
+		log "Packages updating..."
+		glide install
```
migmartri: It will install the new dependencies if glide.lock has changed on container restart.

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '2'
```
migmartri: Moved to the root directory since now it boots both the backend and the frontend in development mode.

src/api/config.yaml
```diff
@@ -0,0 +1,14 @@
+default:
+  cors:
+    allowed-origins:
+      - my-api-server
```
migmartri: We will need to change this value once we know the host of the API.

src/api/glide.yaml
```diff
@@ -7,3 +7,9 @@ import:
 - package: github.com/go-swagger/go-swagger
   version: 0.6.0
 - package: gopkg.in/yaml.v2
+- package: github.com/rs/cors
```
migmartri: I am using an extra package for CORS support.

src/api/pkg/config/config.go
```diff
@@ -0,0 +1,65 @@
+package config
```
migmartri: This package takes care of loading the `config.yaml` file with the values and overrides depending on the ENVIRONMENT env variable

docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    command: "ng serve -h 0.0.0.0"
+  api:
+    tty: true
+    build: ./dev_env/api
+    ports:
+      - 8080:8080
+    volumes:
+      - ./src/api:/go/src/github.com/helm/monocular/src/api
+    environment:
+      - ENVIRONMENT=development
```
migmartri: This env variable enables a very permissive CORS setup.

src/api/pkg/swagger/restapi/configure_monocular.go
```diff
@@ -66,5 +69,26 @@ func setupMiddlewares(handler http.Handler) http.Handler {
 // The middleware configuration happens before anything, this middleware also applies to serving the swagger.json document.
 // So this is a good place to plug in a panic handling middleware, logging and metrics
 func setupGlobalMiddleware(handler http.Handler) http.Handler {
+	handler = setupCorsMiddleware(handler)
+	return handler
+}
+
+func setupCorsMiddleware(handler http.Handler) http.Handler {
```
migmartri: CORS middleware.

src/ui/src/app/chart-search/chart-search.component.ts
```diff
@@ -31,7 +31,7 @@ export class ChartSearchComponent implements OnInit {
   }
 
   searchCharts(q: String): void {
-		this.chartsService.searchCharts(q).then(charts => this.charts = charts);
+		this.chartsService.searchCharts(q).subscribe(charts => this.charts = charts);
```
migmartri: Moved to observables instead of promises.

jackfrancis:
This looks fun, when it's testable could you drop some instructions, I'd like to try it out!

src/api/pkg/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	config := ConfigurationWithOverrides{}
+	res := Configuration{}
+
+	data, err := readConfigFile()
+
+	if err != nil {
+		return res, err
+	}
+
+	err = yaml.Unmarshal([]byte(data), &config)
```
anguslees: (`data` is already `[]byte` here, so the cast is unnecessary)

src/api/pkg/config/config.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	res = mergeConfig(config, currentEnvironment())
+
+	return res, nil
+}
+
+func mergeConfig(conf ConfigurationWithOverrides, env string) Configuration {
+	var defaults Configuration = conf["default"]
+	var custom Configuration = conf[env]
```
anguslees: `custom := conf[env]` is more idiomatic wherever possible (leave it up to the compiler to work out the type)

README.md
```diff
@@ -34,6 +32,17 @@ The API is a golang HTTP server located in `src/api/`.
 
 More API docs are [here](src/api/README.md).
 
-## Status of the Project
+# Running a development environment
```
migmartri: Added top-level documentation on how to run the app.

src/api/pkg/config/config.go
```diff
@@ -0,0 +1,64 @@
...
+func currentEnvironment() string {
+	env := os.Getenv("ENVIRONMENT")
+	if env == "" {
+		env = "production"
+	}
+	return env
+}
+
+func readConfigFile() ([]byte, error) {
+	data, err := Asset("config.yaml")
```
migmartri: The yaml file is actually now bound to the code. See `config/bindata.go`

jackfrancis:
@migmartri I added `coverage.txt` to `.gitignore` in a recent PR, hence the conflict. Should be easy to resolve.

Label this PR as "awaiting review" and I'll test locally and review. Thanks again!

README.md
```diff
@@ -34,6 +32,20 @@ The API is a golang HTTP server located in `src/api/`.
...
+
+We leverage [docker](https://www.docker.com) (via `docker-compose`) to provide a multi-tier setup for development.
+
+Running `docker-compose up` from the root directory will expose:
+
+* API backend endpoint via `http://{your-docker-machine-ip-address}:8080`
+* UI frontend via `http://{your-docker-machine-ip-address}:4200`  
+
+**IMPORTANT**: If your Docker Machine hostname is different than *localhost*, you need to change
```
migmartri: The UI assumes that the backend is under localhost. This is a problem with docker based setups that could be fixed in two ways:

1 - Use something like [Dinghy](https://github.com/codekitchen/dinghy) and enforce static hostnames for frontend and backend.
2 - Offer a way to the developer to change the backend hostname value.

Regarding #1, the problem is that it adds the requirement of running the dinghy-http-proxy in front which might be an overkill for linux and docker for mac users.

In the other hand, it seems that docker for mac uses localhost in the same way than the default Linux does.

Regarding #2, the problems is that angular-cli does not offer a way to extend the [WebPack configuration](https://github.com/angular/angular-cli/issues/1656) so we can not load this Hostname Value based on an environment variable.

So for now and while this changes, we ask the developer to change the hardcoded value in the `config.ts`.

src/api/config/config.go
```diff
@@ -0,0 +1,69 @@
...
+
+func currentEnvironment() string {
+	env := os.Getenv("ENVIRONMENT")
+	if env == "" {
+		env = "production"
+	}
+	return env
+}
+
+func readConfigWithOverrides() configurationWithOverrides {
```
migmartri: Instead of having a yaml file, we just have the structs here.

That removes the requirement of binding the yaml file to the binary file.

src/ui/src/app/config.ts
```diff
@@ -0,0 +1,7 @@
+/* TODO. In the future this value should be read from an env variable
+ and injected by Webpack. But today Webpack configuration extension is not
+ supported in angular-cli https://github.com/angular/angular-cli/issues/1656
+*/
+export const CONFIG = {
+ backendHostname: "http://localhost:8080"
```
migmartri: File that needs to be changed if using a different hostname for the backend.

jackfrancis:
Tested locally, looks good. I agree that we'll revisit some of the hardcoding, but for now this works great!

PR #4, opened 2016-08-26
feat(data): added get repo+chart endpoint

jackfrancis:
- created a `monocular/data` package to facilitate generic data handling
- created a `monocular/data/mocks` package to support mocks for prototyping/testing
- swagger spec updated to include a new `GET /v1/charts/{:repo}/{:chartName}` endpoint
 - added three new data models to accommodate `^--`: `chart`, `chartAttributes`, and `error` (the last of which is a generic error object to return as JSON for non-200 responses)

PR #24, opened 2016-10-05
ref(data): charts data business logic now behind a data.Charts interface

jackfrancis:
Everything is still statically defined for mock data, but this is a graceful path forward toward a functional implementation of a charts data layer.

- the increasingly ugly `mocks.getTestDataWd()` convenience function updated to support mock data retrieval contexts in the root git directory (e.g., to enable `make -C src/api test` and related commands from the `monocular` root dir)
- handlers updated to support the new `Charts` interface type, server configuration likewise dialed in
- tests updated


src/api/data/charts.go
```diff
@@ -8,32 +8,47 @@ import (
 	"github.com/helm/monocular/src/api/pkg/swagger/models"
 )
 
-// GetChart gets the chart associated with the passed-in repo+name
-func GetChart(repo, name string) (models.Resource, error) {
-	// TODO implement actual "get chart" business logic
+// Charts is an interface for managing chart data
+type Charts interface {
```
technosophos: Why are all of these `Get*()`? Couldn't they just be `Chart` (or `Get`), `All`, `AllFromRepo`

src/api/data/charts.go
```diff
@@ -8,32 +8,47 @@ import (
...
+	// will have a GetChart method to retrieve a particular chart from a repo
+	GetChart(repo, name string) (models.Resource, error)
```
technosophos: Why does `GetChart` return a `Resource` (instead of a Chart), and why does it return a `model.Resource` while the others all return `*model.Resource`?

src/api/data/charts.go
```diff
@@ -8,32 +8,47 @@ import (
...
+	// will have a GetAllFromRepo method to retrieve all charts from a repo
+	GetAllFromRepo(repo string) ([]*models.Resource, error)
+	// will have a GetAll method to retrieve all charts from all repos
+	GetAll() ([]*models.Resource, error)
+}
+
+// mockCharts fulfills the Charts interface
+type mockCharts struct{}
```
technosophos: Why hide the implementation of a mock?

src/api/data/charts.go
```diff
@@ -8,32 +8,47 @@ import (
...
+
+// NewMockCharts returns a new mockCharts
+func NewMockCharts() Charts {
+	return &mockCharts{}
+}
+
+// GetChart method for mockCharts
+func (g *mockCharts) GetChart(repo, name string) (models.Resource, error) {
```
technosophos: There are a lot of functions that look like this, but this looks sorta suspect to me. Basically, you match a method signature, run a function with a different name, trap its error, hide the error with a replacement error, and then return the results. It's impossible to recover the actual cause of any given error. This seems really convoluted. Seems to me it could have just been reduced to `return mocks.GetChartFromMockRepo(repo, name)`

technosophos:
I stopped part way through. Honestly, having a hard time following all the layers.

src/api/data/charts.go
jackfrancis: Changed in commit #2.

src/api/data/charts.go
jackfrancis: addressed in commit #4

src/api/data/charts.go
jackfrancis: it's a bit academic, but for now there's now exportable entrance use-case outside of the constructor, so we'll keep the interface struct implementation private

PR #26, opened 2016-10-11
feat(data): adds a data cache, periodic data freshener

jackfrancis:


src/api/data/cache/cache.go
```diff
@@ -0,0 +1,87 @@
...
+	knownRepos []map[string]string
+	allCharts  map[string][]*models.ChartVersion
+	rwm        *sync.RWMutex
+}
+
+// NewCachedCharts returns a new cachedCharts
+func NewCachedCharts(repos []map[string]string) data.Charts {
+	return &cachedCharts{
+		knownRepos: repos,
+		rwm:        new(sync.RWMutex),
```
technosophos: If rwm is not a mutex, you don't need this line since "the zero value for a RWMutex is an unlocked mutex."

src/api/data/cache/cache.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/helm/monocular/src/api/mocks"
+	"github.com/helm/monocular/src/api/swagger/models"
+)
+
+var mockCharts = mocks.NewMockCharts()
+
+type cachedCharts struct {
+	knownRepos []map[string]string
+	allCharts  map[string][]*models.ChartVersion
+	rwm        *sync.RWMutex
```
technosophos: I think you don't want a pointer in this case. You just want a `sync.RWMutex`

arschles: this will be fine as-is if you change to a `sync.RWMutex`, but might wanna leave it a pointer as a defensive measure.

if you accidentally pass an instance of this struct by value like this:

```go
func (c cachedCharts) doSomething() {...}
```

your code will make a copy of the `sync.RWMutex` and render the mutex semantically incapable of doing mutual exclusion as you'd expect (`go vet` would complain too). here's a trivial example: https://play.golang.org/p/dendtURMEP

n.b. - there's no technical reason to remove the pointer right now, and the compiler would likely also complain for other reasons if you added a method without a pointer receiver. I'm being paranoid about this because it's a concurrency issue

src/api/data/cache/auto-refresher.go
```diff
@@ -0,0 +1,46 @@
...
+		chartsImplementation: chartsImplementation,
+		frequency:            frequency,
+		name:                 name,
+	}
+}
+
+// Do implements the Periodic interface
+func (r *refreshChartsData) Do() error {
+	err := r.chartsImplementation.Refresh()
+	if err != nil {
```
arschles: combine this and the above into

```go
if err := r.chartsImplementation.Refresh(); err != nil {
  return err
}
```

src/api/data/cache/cache.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// NewCachedCharts returns a new cachedCharts
```
arschles: prob make this say 'returns a new Charts implementation'

src/api/data/cache/cache.go
```diff
@@ -0,0 +1,87 @@
...
+func NewCachedCharts(repos []map[string]string) data.Charts {
+	return &cachedCharts{
+		knownRepos: repos,
+		rwm:        new(sync.RWMutex),
+		allCharts:  make(map[string][]*models.ChartVersion),
+	}
+}
+
+// ChartFromRepo method for cachedCharts
```
arschles: this doc string is not meaningful. if it's an interface implementation, probably should just say something like "ChartFromRepo is the interface implementation for data.Charts. it behaves this way: ....."

arschles: same for the below

src/api/data/cache/cache.go
```diff
@@ -0,0 +1,87 @@
...
+	defer c.rwm.RUnlock()
+	var allCharts []*models.Resource
+	for _, repo := range c.knownRepos {
+		for repoName := range repo {
+			for _, chart := range c.allCharts[repoName] {
+				resource := helpers.MakeChartResource(chart, repoName)
+				allCharts = append(allCharts, resource)
+			}
+		}
+	}
```
arschles: any way you can reduce the time complexity here? maybe by processing each repo concurrently? I'm assuming that there are going to be a lot of repos at some point; if that's not gonna be the case, ignore!

src/api/data/charts.go
```diff
@@ -10,4 +10,6 @@ type Charts interface {
 	AllFromRepo(repo string) ([]*models.ChartVersion, error)
 	// will have a All method to retrieve all charts from all repos
 	All() ([]*models.Resource, error)
+	// will have a Refresh method to freshen charts data
```
arschles: Can you change this doc string to say "Refresh freshens ...."

src/api/jobs/jobs.go
```diff
@@ -0,0 +1,46 @@
...
+	// responsible for executing it every Frequency() thereafter
+	Do() error
+	Frequency() time.Duration
+	Name() string
+}
+
+// DoPeriodic calls p.Do() once, and then again every p.Frequency() on each element p in pSlice.
+// For each p in pSlice, a new goroutine is started, and the returned channel can be closed
+// to stop all of the goroutines.
+func DoPeriodic(pSlice []Periodic) chan<- struct{} {
```
arschles: is it necessary to have the `Periodic` interface? it has 3 functions defined on it, and could be simplified to a `func() (string, time.Duration, func() error)`. just an idea - would probably save some boilerplate

arschles: also, might be good to abstract away the returned `chan<- struct{}` into a `func()`. user can then call it to close all the goroutines instead of dealing with the channel implementation detail

src/api/swagger/restapi/configure_monocular.go
```diff
@@ -23,7 +25,19 @@ func configureFlags(api *operations.MonocularAPI) {
...
+		},
+		map[string]string{
+			"incubator": "https://github.com/kubernetes/charts/tree/master/incubator",
+		},
+	}
+	chartsImplementation := cache.NewCachedCharts(repos)
+	freshness := time.Duration(3600) * time.Second
+	periodicRefresh := cache.NewRefreshChartsData(chartsImplementation, freshness, "refresh-charts")
+	toDo := []jobs.Periodic{periodicRefresh}
+	_ = jobs.DoPeriodic(toDo)
```
arschles: omit the `_ = ` if you're not gonna do anything with the returned value

arschles:
@jackfrancis I'm completely unfamiliar with what monocular does, so forgive if this is an irrelevant question: where is this cache ultimately going to be built from? my concern is that, if this is a server among a bunch of replicas, it will hold a bunch of state (the cache) that could easily become stale but not invalidated

jackfrancis:
@arschles the cache will be aggregate data from a federation of helm repository servers, each of which will hold an index representation of charts.

src/api/data/cache/cache.go
jackfrancis: It's a good note, we don't anticipate lots of repos, but the code will benefit from a note that suggests that this won't scale well.

arschles:
@jackfrancis got it. that might make https://github.com/helm/monocular/pull/26/files#r83332114 important sometime in the future. anyway, this LGTM for now

PR #27, opened 2016-10-17
ref(data): incorporates latest changes to index.yml spec

jackfrancis:
Of especial notice:

- data model property changes: 'checksum' becomes 'digest', 'url' becomes a 'urls' array, a new 'sources' array
- helper conveniences updated to handle new data models
- mock data updated to reflect new index.yml spec


jackfrancis:
Here is a relevant commit that documents the updates to the `index.yaml` format:

https://github.com/kubernetes/helm/commit/491dc325521976046a7738250802efcc32e5c400

Issue #33, opened 2016-10-18
API - latest versions link has incorrect route string


jackfrancis:
Additionally, it works inconsistently. The "latest" link should really only appear on the following route endpoint:

- `/{:apiVersion}/charts/{:repoName}/{:chartName}`

It doesn't make sense to have a "latest" link on the following endpoints:

- `/{:apiVersion}/charts`
- `/{:apiVersion}/charts/{:repoName}`
- `/{:apiVersion}/charts/{:repoName}/{:chartName}/versions`
- `/{:apiVersion}/charts/{:repoName}/{:chartName}/versions/{:chartVersion}`

Issue #30, opened 2016-10-18
API - /charts/{repo}/{chartName} <=> /charts/{repo}/{chartName}/versions


migmartri:
Apparently based on my tests `http://localhost:8080/v1/charts/stable/redmine` and `http://localhost:8080/v1/charts/stable/redmine/versions` return both the same result and the same kind of object (chart).

Is this a desired behavior? Maybe for the /stable/redmine/versions API endpoint, we should change the type to `chartVersion`, and for the `/stable/redmine` case, remove the version specific attributes as suggested here https://github.com/helm/monocular/issues/29.

What's your take on this?

jackfrancis:
They should not return the same type of data. This example route:

- `http://localhost:8080/v1/charts/stable/redmine`

... should return a non-version-specific representation of the chart, with a link to the "latest" version

This example route:

- http://localhost:8080/v1/charts/stable/redmine/versions

... should return a collection of all specific chart+version data representations, but without the "latest" link present.

jackfrancis:
See #29 for more info. But, in summary, I think the easiest path forward is to in fact include the same kind of versioned data for all endpoints, with the exception that the `/v1/charts/{:repo}/{:chart}` endpoint will include a `links.latest` canonical link to the returned version.

Issue #29, opened 2016-10-18
API - /charts/{repo}/{chart-name} observations


migmartri:
If you access for example to `http://localhost:8080/v1/charts/stable/redmine` you get the following output.

```
{
data: {
  attributes: {
    created: "2016-10-06T16:23:20.499814565-06:00",
    description: "A flexible project management web application.",
    digest: "99c76e403d752c84ead610644d4b1c2f2b453a74b921f422b9dcb8a7c8b559cd",
    home: "https://k8s.io/helm",
    name: "redmine",
    repo: "stable",
    sources: [
      "https://github.com/kubernetes/charts/redmine"
   ],
   urls: [
     "https://storage.googleapis.com/kubernetes-charts/redmine-0.0.1.tgz"
  ]
  },
  id: "stable/redmine",
  links: {
    latest: "/v1/charts/stable/redmine/0.0.1"
  },
  type: "chart"
 }
}
```

#### Latest Link
For example the latest link points to `/v1/charts/stable/redmine/0.0.1` but I am wondering if it should include the `/versions` namespace, i.e /v1/charts/stable/redmine/versions/0.0.1

#### Digest

Shouldn't this route return a non-version Chart resource. If that's the case, shouldn't we remove digest and the Urls?

jackfrancis:
I think these issues should address all of this:

- https://github.com/helm/monocular/issues/33
- https://github.com/helm/monocular/issues/36

The only version info that the general chart route endpoint should deliver is a link to the "latest" version endpoint.

jackfrancis:
I'm recommending that we keep the fully populated, versioned metadata inside the data response for `/v1/charts/{:repo}/{:chart}`. Everything will be in there to be consumed when it's appropriate to source data from a "get latest chart version" authoritative endpoint.

We'll also include the canonical link for that version in the `links` object, in its `latest` property.

PR #40, opened 2016-10-18
ref(data): tweaks to versioned data representation

jackfrancis:
changes:

- now all chart representations include a "version" attribute that represents the version string of the returned, versioned chart
- "links.latest" canonical link is only included in the "get latest version" route endpoint (`/v1/charts/{:repo}/{:chart}`)


jackfrancis:
closes #29 
closes #30 
closes #33

src/api/handlers/charts.go
```diff
@@ -20,6 +20,7 @@ func GetChart(params operations.GetChartParams, c data.Charts) middleware.Respon
 		return notFound(chartResourceName)
 	}
 	chartResource := helpers.MakeChartResource(chart, params.Repo)
+	helpers.AddLatestLinks(chartResource, *chart.Version)
```
migmartri: I thought that the latest link was going to be accessible via `/charts` and `/charts/{repo}` which return a non-versioned Chart resource no?

Basically only removing it from /charts/{repo}/{chartName}/versions

jackfrancis: The API doesn't currently return non-versioned chart resources. We should consider doing that separately:

https://github.com/helm/monocular/issues/41

PR #46, opened 2016-10-20
ref(api): distinct typed API responses

jackfrancis:
Distinct "chart", "chartVersion" and "chartVersionLatest" API response types.

- returning a "chart" type for canonical repo+chart endpoint (e.g., `/{:apiVersion}/charts/{:repo}/{:chart}`, including a "relationship" property that itself contains a "chartVersionLatest" data property whose metadata describes the latest version of that repo+chart
- returning an array of "chart" types for the "list all charts" and "list all charts in a repo" API endpoints, including a canonical link with each "chart" entry
- returning a "chartVersion" type for the version-specific repo+chart API endpoint
- returning an array of "chartVersion" types for the "list all chart versions in a repo" API endpoint

Also, important: this includes new mock data! (copied from the current `http://storage.googleapis.com/kubernetes-charts/index.yaml`)


jackfrancis:
screen shot 2016-10-19 at 5 23 38 pm
"chart" type array

jackfrancis:
screen shot 2016-10-19 at 5 24 37 pm
"chart" type response with canonical link

jackfrancis:
screen shot 2016-10-19 at 5 25 38 pm

"chartVersion" response (e.g. `/v1/charts/stable/redmine/versions/0.3.3`)

jackfrancis:
screen shot 2016-10-19 at 5 28 19 pm

"chartVersion" type array (e.g., `/v1/charts/stable/redmine/versions`)

src/api/mocks/testdata/repo-incubator.yaml
```diff
@@ -1,25 +1,40 @@
 apiVersion: v1
```
prydonius: The stable repo index has been updated with the correct order so you may want to update it, but not absolutely necessary for Monocular.

prydonius:
Great work @jackfrancis, though I'm not sure about adding the chartVersionLatest type and having that differ from chartVersion - I think this may be diverging a little too much.

As a simplification, I suggest that we describe two resources:

```
chart:
  description
  home
  keywords
  maintainers
  name
  repo
  sources

chartVersion:
  created
  digest
  urls
  version
```

So the `chart` type, is everything in the Chart.yaml that is un-versioned metadata about the Chart, and the metadata here should only reflect what's in latest version's Chart.yaml. This is what we already have, which is great! `chartVersion` on the other hand, only holds information about where to find a specific version of a Chart, and not any metadata about what the Chart is. So this would involve removing the duplicated content from the current `chartVersion` and make it match `chartVersionLatest`

I also think when requesting both resources, they should hold a relationship to each other. A rough idea is something like this:

GET /v1/charts/stable/redmine
```json
{
  "data": {
    "attributes": {
      "description": "A flexible project management web application.",
      "home": "https://k8s.io/helm",
      "keywords": ...,
      "maintainers": ...,
      "name": "redmine",
      "repo": "stable",
      "sources": [
        "https://github.com/kubernetes/charts/redmine"
      ],
    },
    "id": "stable/redmine",
    "relationships": {
      "latestChartVersion": {
        "data": {
          "created": "2016-10-06T16:23:20.499814565-06:00",
          "digest": "99c76e403d752c84ead610644d4b1c2f2b453a74b921f422b9dcb8a7c8b559cd",
          "urls": [
            "https://storage.googleapis.com/kubernetes-charts/redmine-0.0.1.tgz"
          ],
          "version": "0.0.1"
        },
        "links": {
          "related": "/v1/charts/stable/redmine/versions/0.0.1"
        }
        "type": "chartVersion" // I think as long as type is here, we can use latestChartVersion in the key
      }
    },
    "type": "chart"
  }
}
```

GET /v1/charts/stable/redmine/versions/0.0.1
```json
{
  "data": {
    "attributes": {
      "created": "2016-10-06T16:23:20.499814565-06:00",
      "digest": "99c76e403d752c84ead610644d4b1c2f2b453a74b921f422b9dcb8a7c8b559cd",
      "urls": [
        "https://storage.googleapis.com/kubernetes-charts/redmine-0.0.1.tgz"
      ],
      "version": "0.0.1"
    },
    "relationships": {
      "chart": {
        "data": {
          "description": "A flexible project management web application.",
          "home": "https://k8s.io/helm",
          "name": "redmine",
          "repo": "stable",
          "sources": [
            "https://github.com/kubernetes/charts/redmine"
          ]
        },
        "links": {
          "related": "/v1/charts/stable/redmine"
        },
        "type": "chart"
      }
    },
    "type": "chartVersion"
  }
}
```

This way, we can solve the issue of presenting the Chart metadata when requesting a specific version, but still keep the concerns of the `chart` and `chartVersion` types separate, and without introducing anymore objects. I really think this approach will simplify the API resources a lot. What do you guys think?

jackfrancis:
@prydonius that should work, I'll do another iteration incorporating your comments

jackfrancis:
@prydonius thanks for the feedback. I'd like to propose the following:

- Keep your proposed distinction between `chart` and `chartVersion`, but change `chartVersion` to `chartPackage` (or something like that). The reason for this is that we already have fairly significant code dependencies on a thing called a `ChartVersion` that represents a composition of `Chart` data + `Version` data.
- `/v1/charts/stable/redmine` response will include a relationship to a `chartPackage`, which will represent the latest published version
- `/v1/charts/stable/redmine/versions/0.0.1` response will be a `chartPackage` with a relationship to a `chart` as you propose
- `/v1/charts/stable/redmine/versions` response will be an array with items of type `chartPackage`

A question: do we want the items in the final bullet-point above to also include relationships to the main `chart` route?

prydonius:
@jackfrancis hmm, I see. The issue with `chartPackage` is the route is `.../versions` which may seem a little confusing. Would it make more sense to rename the `ChartVersion` struct to `ChartPackage`, as that would be more true to the description of Chart data + Version data - or is that a mammoth task in itself?

prydonius:
> A question: do we want the items in the final bullet-point above to also include relationships to the main chart route?

I think for brevity we shouldn't include relationships in list responses, we could choose to use links here though.

jackfrancis:
@prydonius I think `ChartPackage` is a less accurate description than `ChartVersion` of this:

```
{
  attributes: {
    created: "2016-10-19T00:03:14.024557743Z",
    description: "One of the most versatile open source content management systems.",
    digest: "c56fc55b93b0dead65af7b81bbd54befd5115860698ca475baa5acb178a12e5a",
    home: "http://www.drupal.org/",
    keywords: [
      "drupal",
      "cms",
      "blog",
      "http",
      "web",
      "application",
      "php"
    ],
    maintainers: [
      {
        email: "containers@bitnami.com",
        name: "Bitnami"
      }
    ],
    name: "drupal",
    repo: "stable",
    sources: [
      "https://github.com/bitnami/bitnami-docker-drupal"
    ],
    urls: [
      "http://storage.googleapis.com/kubernetes-charts/drupal-0.3.0.tgz"
    ],
    version: "0.3.0"
  },
  id: "stable/drupal:0.3.0",
  type: "chartVersion"
}
```

I would prefer to change the `versions` routes to `packages` instead if we want to be semantically correct in terms of response data type. That would mean:

- `/v1/charts/stable/redmine/packages/0.0.1`
- `/v1/charts/stable/redmine/packages`

jackfrancis:
O.K., so we'll go with the most recent proposal with @prydonius suggestions:

- `/v1/charts/stable/redmine` response will include a relationship to a `chart`, which will include a relationship link to the latest version of type `chartVersion`
- `/v1/charts/stable/redmine/versions/0.0.1` response will be a `chartVersion` with a relationship to a `chart`
- `/v1/charts/stable/redmine/versions` response will be an array with items of type `chartVersion`
 - each entry will include a link to the canonical chart entry

Internal use of the `ChartVersion` name will be changed to `ChartPackage`.

migmartri:
Great, this is taking good shape!

 Please see my comments inline in the attached image which summarizing say:

* Double check the correct use of links  (canonical vs related vs self)
* Rename chartVersion to latestChartVersion
* Maybe add a link to the `chart` from the version (low prior)
* My take regarding the relationship is to add it always, regardless if you are asking for an specific chart or not. I know that we are doing it this way to reduce the payload size, but in my opinion, the payload will never get too big since `v1/charts` will return a grouped version of charts (not versions), and we will need to set pagination eventually. 

Adding the relationship always, will allow us in the UI to get the latestVersion number when we list the charts, it will also simplify the business logic of the API.

 In any case, none of my comments are a blocker and we can iterate later. 

Thanks!

selection_003

prydonius:
> My take regarding the relationship is to add it always, regardless if you are asking for an specific chart or not. I know that we are doing it this way to reduce the payload size, but in my opinion, the payload will never get too big since v1/charts will return a grouped version of charts (not versions), and we will need to set pagination eventually.

After talking to @migmartri, I wanted to make this clearer and lay out what we think the links and relationships should look like for each route.

- /charts, /charts/:repo => for each chart
  - relationships
    - latest chart version
  - links
    - self/canonical
    - all chart versions
- /charts/:repo/:chart (same payload as an item in previous route)
  - relationships
    - latest chart version
  - links:
    - self/canonical
    - all chart versions
- /charts/:repo/:chart/versions => for each version
    - links:
      - self/canonical
      - chart

@jackfrancis what do you think?

jackfrancis:
New iteration is ready for review with the recommended changes, minus the "all chart versions" links decoration (punting on that for now).

migmartri:
I have noticed that in the URL of the type `v1/charts/REPO/CHART/versions`, we add a `chart` relationship with the expanded data (and repeated data). 

In the specification that Adnan laid out, I read that in that route, the chart is linked from inside a `links` payload not from a relationship.

In any case, I do not think that it is a big deal and we should carry on and close this PR. So I am lgtm :)

Thanks!

PR #47, opened 2016-10-25
feat(api): a basic search implementation

jackfrancis:
- new API endpoint at {:apiVersion}/charts/search that accepts a single query parameter "name"
- "name" query parameter does a substring search against all avail charts' name properties, e.g., /v1/charts/search?name=drupal
- "name" parameter is required (a GET against the URL w/out the "name" param results in a 422 HTTP response)
- search queries that result in no results found return an HTTP 200 w/ a null data response


prydonius:
Tried it out and this lgtm!

Future improvements (should probably make these into PRs)
[] For simplicity, accept a single param (`query`) that search across multiple fields
[] Include keywords in search (a search for database should bring up MySQL, MariaDB, PostgreSQL, etc.)
[] Search descriptions and READMEs
[] Full text search implementation across READMEs and descriptions

jackfrancis:
@prydonius https://github.com/helm/monocular/issues/49

Issue #42, opened 2016-10-19
Ingest Chart data from real Chart repositories


prydonius:
Currently the API returns mocked chart data, we should change this behaviour to fetch from real repos.

There should be a configuration file that lists the Chart repositories to fetch from, by default it will point to the stable and incubator repositories (http://storage.googleapis.com/kubernetes-charts and http://storage.googleapis.com/kubernetes-charts-incubator)

PR #50, opened 2016-10-25
feat(api): now sourcing real repo data in data.cache

jackfrancis:
This is a candidate for hiding behind an interface at some point. For now, we simply include with the cache implementation of data.Charts.Refresh() some http code to fetch real repo data.

### DIFF
--- a/src/api/data/cache/cache_test.go
+++ b/src/api/data/cache/cache_test.go
@@ -4,13 +4,13 @@ import (
 	"testing"
 
 	"github.com/arschles/assert"
+	"github.com/helm/monocular/src/api/data"
 	"github.com/helm/monocular/src/api/data/helpers"
-	"github.com/helm/monocular/src/api/mocks"
 	"github.com/helm/monocular/src/api/swagger/restapi/operations"
 	"github.com/helm/monocular/src/api/testutil"
 )
 
-var chartsImplementation = mocks.NewMockCharts()
+var chartsImplementation = getChartsImplementation()
 
 func TestCachedChartsChartFromRepo(t *testing.T) {
 	// TODO: validate chart data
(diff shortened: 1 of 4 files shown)


technosophos:
Does this mean that the tests now run against live data?

src/api/swagger/restapi/server_test.go
```diff
@@ -187,3 +188,17 @@ func urlPath(ver string, remainder ...string) string {
 func httpGet(s *httptest.Server, route string) (*http.Response, error) {
 	return http.Get(s.URL + "/" + route)
 }
+
+func getChartsImplementation() data.Charts {
```
migmartri: Would it be possible to reuse this method with the one in `cache_test.go`?

src/api/data/cache/cache_test.go
```diff
@@ -90,3 +74,17 @@ func TestCachedChartsRefresh(t *testing.T) {
 	err := chartsImplementation.Refresh()
 	assert.NoErr(t, err)
 }
+
+func getChartsImplementation() data.Charts {
```
migmartri: I am wondering if we should stub these repositories (removing the external dependency) so we remove the possibility of getting different test results.

It should speed up the tests as well no?

jackfrancis:
fixes #42

jackfrancis:
@migmartri pushed a follow-up commit to use mock charts data in `cache` tests. I have created #52 to address rationalizing the way we handle real and mock data in prod vs test scenarios.

jackfrancis:
O.K., I've got project transition brain. This is the actual issue that will enable us to test the `data.cache` package with mock data:

https://github.com/helm/monocular/issues/53

technosophos:
Understanding that the cache tests are backed by live data, and that we'll need to fix that up at a later point, I'm good with merging this. Today is @jackfrancis 's last official day on this project, so I'd rather see this merged than leave it hanging.

prydonius:
Woot! Awesome work @jackfrancis, we'll miss you :(

jackfrancis:
See you at kubecon... :)

PR #55, opened 2016-11-11
Home view design implementation

Angelmmiguel:
I implemented the Home view based on the v2 design. To complete this implementation I removed `bootstrap-loader` and added two new libraries:

* [angular/material2](angular/material2): Material Design components for Angular2.
* [zellwk/mappy-breakpoints](https://github.com/zellwk/mappy-breakpoints): SASS helper to work with breakpoints.

The final implementation is very similar to the design. I added Material2 library to the project, but all components except the input in the header are custom. I tried to use the material UI components, but they are to specific and it's difficult to extend them. For example, I needed to add several CSS lines to set the color of the line and placeholder of the input.

I think we can continue with this library now, but if we implement more views without material components, in my opinion, the best option is to remove it. I know it's better for contributing, but the style of Material Design is very invasive and I think Monocular should have it's own personality.

## Palette

For the color palette, I selected the following color variations. You can check it in [Material/color](https://material.google.com/style/color.html).

* **Primary**: Blue
* **Secondary**: Purple
* **Warn**: orange

## Chart cards

I updated the design of the cards just to adapt the description text. I added short descriptions to the design, but some of them might be bigger. However, the text is limited to 80 characters. About the colors, I know there are too many blue variations in the cards, but remember the logo will depend on the chart.

## Screenshots

Header

Searching

Chart list

Footer

## Responsive

Home view across several resolutions


Angelmmiguel:
**PS**: I removed the capitalization of chart titles.

migmartri:
LGTM! Merging

PR #56, opened 2016-11-16
Rollback default hostname

migmartri:


Angelmmiguel:
That was my fault, sorry!

Issue #63, opened 2017-01-09
UI not loading


migmartri:
The latest version of the code seems to raise the following error getting stuck in the loading page.

```
snack-bar-container.js:4Uncaught TypeError: Cannot read property 'prototype' of undefined
    at __extends (snack-bar-container.js:4)
    at snack-bar-container.js:28
    at Object.<anonymous> (snack-bar-container.js:122)
    at __webpack_require__ (bootstrap 9464e2d…:52)
    at Object.<anonymous> (snack-bar-ref.js:70)
    at __webpack_require__ (bootstrap 9464e2d…:52)
    at Object.<anonymous> (menu.js:40)
    at __webpack_require__ (bootstrap 9464e2d…:52)
    at Object.<anonymous> (icon.js:264)
    at __webpack_require__ (bootstrap 9464e2d…:52)
```

![selection_047](https://cloud.githubusercontent.com/assets/24523/21778783/ad23a5d0-d659-11e6-856b-5a9b28e5e77f.png)

I am using the bundled docker-compose file and accessing the UI via localhost:4200

It might be related to the changes introduced here https://github.com/helm/monocular/issues/59

Angelmmiguel:
@migmartri did you install the latest modules? You can install them with `npm install`

Angelmmiguel:
I didn't update the version of any package and the problem is related with an `undefined` property in the [material2](https://github.com/angular/material2) library. Other users reported the same issue in material2 repository:

https://github.com/angular/material2/issues/2468
https://github.com/angular/material2/issues/2418

I'm going to check it 👍

migmartri:
Hi @Angelmmiguel 

I can confirm that your latest update in the package.json fixes the issue.

Thanks,
Miguel

PR #64, opened 2017-01-11
Upgrading the material2 package version

Angelmmiguel:
Upgrading the version of material package fixes the bundle problem. However, this package still uses a beta package (5.0.0-beta.12)


Angelmmiguel:
This PR solves #63

Issue #32, opened 2016-10-18
Frontend - Use version specific Readme file 


migmartri:
Currently, the client uses a mocked version (`src/ui/src/assets/mock_readme.md`) of a Readme file in all the charts, we must change this to use the ChartVersion specific Readme content exposed via the monocular API. 

Note: This issue is blocked until [this other issue](https://github.com/helm/monocular/issues/31) is solved.

Issue #31, opened 2016-10-18
API - Expose ChartVersion Readme through chart-specific version API


migmartri:
#### Readme API exposure

The UI requires the retrieval of a README file content in order to render it.

Since the Readme file content can change from version to version, its content should come in the chartVersion API call `/charts/{repo}/{chartName}/versions` and `/charts/{repo}/{chartName}/versions/{version}`

 Maybe we should not add the readme file content to the listing of ChartVersions (`/charts/{repo}/{chartName}/versions`) in order to not overload the API call. 
#### Readme importation

As discussed [here](https://github.com/kubernetes/helm/issues/1098), the Readme file for now is not going to be exposed in the index.yaml file but it is part of the `tgz` file, so we would need to do some "extra work" during chart versions ingestion.

Issue #43, opened 2016-10-19
Icons processing and storage


migmartri:
As explained [here](https://github.com/helm/monocular/issues/37), the Charts index contains an icon attribute which represents an iconURL.

For the initial version of Monocular, it is fine that we show the images pointed by those external links but for future versions we should think of:

At importation time:
- Get the images, **process it** to our desired format and aspect ratio.
- Store the resulting image in a known persistent location.

Then the Monocular backend will expose these images from the new location. This will avoid having wrong images aspect ratios and possible attacks/outages because of the image hotlink.

PR #74, opened 2017-02-01
Download, expose and use Chart Readme files

migmartri:
## TLDR:

* Backend: Fetches and exposes specific chartVersion Readme.md files via the API https://github.com/helm/monocular/issues/31
* Frontend: Uses this new readme file https://github.com/helm/monocular/issues/32

## Importation
* During repository fetching, we download and store in the filesystem ($HOME/repo-data) the Readme files for those versions.  We must persist that directory as a volume, otherwise subsequent `refresh` calls will be expensive.

* This mechanism can be useful if we decide to store icons on importation time so we can process them and avoid possible issues related to hotlinking https://github.com/helm/monocular/issues/43 
selection_092

## Exposure
* Readme files are exposed via `.../version/{version}/readme` API endpoint following open-api.org convention
selection_096
selection_097

Closes https://github.com/helm/monocular/issues/31

## UI

selection_098

Closes  https://github.com/helm/monocular/issues/32

## Caveats

* I know that testing and coverage could be improved but I'd prefer to land this asap so I do not block the work happening in the UI.
* I'm not familiar with Go so I tried to follow existing code conventions, in any case it is likely that I am making basic mistakes so harsh feedback is welcome :)


src/api/data/cache/cache.go
```diff
@@ -143,6 +144,13 @@ func (c *cachedCharts) Refresh() error {
 			c.allCharts[repoName] = []*models.ChartPackage{}
 			for _, chart := range charts {
 				chart.Repo = repoName
+				// Extra files. Skipped if the directory exists
+				if _, err := os.Stat(chartDataDir(chart)); os.IsNotExist(err) {
```
migmartri: We download the files if cache does not exist for that specific version.

src/api/handlers/charts.go
```diff
@@ -41,6 +42,28 @@ func GetChartVersion(params operations.GetChartVersionParams, c data.Charts) mid
...
+func GetChartVersionReadme(params operations.GetChartVersionReadmeParams, c data.Charts) middleware.Responder {
+	// We check that the chart exists before load the readme
+	chartPackage, err := c.ChartVersionFromRepo(params.Repo, params.ChartName, params.Version)
+	if err != nil {
+		log.Printf("data.Charts.ChartVersionFromRepo(%s, %s, %s) error (%s)", params.Repo, params.ChartName, params.Version, err)
+		return notFound(ChartVersionResourceName)
+	}
+
+	filename := "README.md"
+	readmeContent, err := cache.ReadFromCache(chartPackage, filename)
```
migmartri: Using the convenience method created in `cache package`

src/api/swagger.yml
```diff
@@ -88,6 +88,32 @@ paths:
...
+          required: true
+        - name: version
+          in: path
+          type: string
+          required: true
+      responses:
+        200:
+          description: Readme.md file content
+          schema:
+            $ref: "#/definitions/resourceDataString"
```
migmartri: New resource to simplify response. It includes `data` root. I need to add `type` and `id` to the payload to make it fully complaint with the jsonapi spec

Angelmmiguel:
Are we going to parse and render the README in Frontend @migmartri??

migmartri:
> Are we going to parse and render the README in Frontend @migmartri??
@Angelmmiguel, retrieval, parsing and rendering will be UI concern. 

We will need to remove the mocked Readme file we have in the UI and make a call to the provided version specific API endpoint instead.

src/api/data/cache/cache-helper.go
```diff
@@ -0,0 +1,212 @@
+package cache
```
prydonius: This filename isn't really idiomatic, I suggest dropping the cache prefix and going with `helpers.go` or `tar_helpers.go` or `cache_helper.go`.

src/api/handlers/charts.go
technosophos: That looks good.

src/api/data/cache/cache-helper.go
```diff
@@ -0,0 +1,212 @@
...
+		return err
+	}
+
+	if !tarballExists(chart) {
+		if err := downloadTarball(chart); err != nil {
+			return err
+		}
+	}
+
+	if err := extractFilesFromTarball(chart); err != nil {
```
technosophos: tiny nit: you could just simplify this to `return extractFilesFromTarball(chart)`

src/api/data/cache/cache-helper.go
```diff
@@ -0,0 +1,212 @@
...
+
+	fmt.Printf("Downloading metadata from %s\n", source)
+	// Download tarball
+	resp, err := http.Get(source)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+	_, err = io.Copy(out, resp.Body)
+	if err != nil {
```
technosophos: similar, here you could just `return err`.

PR #75, opened 2017-02-01
Retrieve the icon of the chart from the API

Angelmmiguel:
Now, the component that represents the basic Chart information uses the Icon retrieved from the API. If the icon attribute is undefined, it displays the placeholder icon. 

However, we must manage the icons we are showing. Most of them don't follow any convention so they are misplaced. To center them, I tried to add the images as `background`. Angular prevents me to write the value into the HTML directly due to XSS problems. In fact, it makes sense because the icon property is an string so it can contains malicious code. Also, some icons may not be available.

Do you think we can cache and process the icons that we show in the UI?

List of charts

In addition, I refactored the location of `models` and `services`. I moved those classes to `shared` folder.


migmartri:
@Angelmmiguel, downloading the icons and serving them from the backend is doable once we add the code that saves the readme file https://github.com/helm/monocular/pull/74

It might take some time though. We can iterate on it later.

Angelmmiguel:
👍  @migmartri

Angelmmiguel:
I'm going to merge the code. We will improve this solution when the backend provides the icons.

PR #76, opened 2017-02-02
feat(*): add helm chart

prydonius:


migmartri:
A couple of things that I'd add to this PR

- [ ] Rebase master
- [ ] build/deploy scripts
- [ ] Be sure that the api endpoint configuration in the client is modified only for the production environment so it does not break the dev env.

helm/monocular/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+description: A Helm chart for Kubernetes
```
migmartri: Can you update the description ;)

Take something from the readme.

helm/monocular/values.yaml
```diff
@@ -0,0 +1,39 @@
...
+  replicaCount: 2
+  image:
+    repository: gcr.io/helm-ui/monocular-api
+    tag: latest
+    pullPolicy: Always
+  service:
+    name: monocular-api
+    type: NodePort
+    externalPort: 80
+    internalPort: 8081
```
migmartri: You should be able to run the server with port 80 with the `PORT` env variable.

In any case, I do not think that it is needed.

src/ui/rootfs/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM bitnami/nginx
+
+COPY dist/ /app
```
migmartri: can you add a comment with the process that needs to be previously done?

`ng build --prod ....`

helm/monocular/values.yaml
prydonius: I would rather have the container run a non-privileged port, even though it starts as root currently.

PR #95, opened 2017-02-07
Update the main theme color of the site

Angelmmiguel:
Also, I fixed some minor improvements in versions and repos of the application cards.


Angelmmiguel:
You can see the color palette here: http://codepen.io/angelrb/full/QdBYqN/

Issue #88, opened 2017-02-04
Make charts listing order deterministic


migmartri:
In the UI, the order of the charts seem to be defined in a random order probably generated at importation time.

We need to make it deterministic, maybe setting a default sorting by name.

This is specially a problem if we enable replication in the backend in which case we will get different order depending on the backend responding the request.

PR #97, opened 2017-02-07
Sort by attributes.Name

migmartri:
* Created sorter interface implementation to sort by Name and applied it to `/charts` and `/charts/:repo`
* This sorter implementation will be probably extended to support two dimensions sorting, first by repository.

Closes https://github.com/helm/monocular/issues/88

### DIFF
--- a/src/api/data/cache/sorting.go
+++ b/src/api/data/cache/sorting.go
@@ -1,0 +1,22 @@
+package cache
+
+import "github.com/helm/monocular/src/api/swagger/models"
+
+/*
+This interface implementation will be
+extended to support sorting by repostory first
+*/
+
+// ByName sorts by Name property
+type ByName []*models.ChartPackage
+
+// Sorting by name
+func (c ByName) Len() int {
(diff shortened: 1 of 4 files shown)


src/api/data/cache/cache_sorting.go
```diff
@@ -0,0 +1,22 @@
+package cache
```
prydonius: rename to `sorting.go`

src/api/data/cache/sorting.go
```diff
@@ -0,0 +1,22 @@
...
+type ByName []*models.ChartPackage
+
+// Sorting by name
+func (c ByName) Len() int {
+	return len(c)
+}
+func (c ByName) Swap(i, j int) {
+	c[i], c[j] = c[j], c[i]
+}
+func (c ByName) Less(i, j int) bool {
```
migmartri: I decided to implement the interface because it is likely that we will have a more complex two dimensions sorting (repo - name)

PR #98, opened 2017-02-08
Icons processing and exposure

migmartri:
* Moved helper files to custom package (charthelper)
* Downloads and process icons
* Added static files middleware under `/assets` (we could eventually put Nginx to serve this path directly instead)
* Remove Readme file endpoint and exposed now via the static middleware.
* Exposes processed icons via the static middleware
* Icons and README static paths are exposed via the API 

Chart endpoint:
selection_110

* The backend now serves the static files directory that contains those files.

selection_109

Closes https://github.com/helm/monocular/issues/43


src/api/data/cache/cache_test.go
```diff
@@ -101,13 +107,14 @@ func TestCachedChartsRefreshErrorPropagation(t *testing.T) {
...
 
 func TestCachedChartsRefreshErrorDownloadingPackage(t *testing.T) {
-	chartDataExistsOrig := chartDataExists
-	defer func() { chartDataExists = chartDataExistsOrig }()
-	chartDataExists = func(chart *models.ChartPackage) (bool, error) { return false, nil }
-	downloadAndExtractChartTarballOrig := downloadAndExtractChartTarball
-	defer func() { downloadAndExtractChartTarball = downloadAndExtractChartTarballOrig }()
-	knownError := errors.New("error on downloadAndExtractChartTarball")
-	downloadAndExtractChartTarball = func(chart *models.ChartPackage) error {
+	ChartDataExistsOrig := charthelper.ChartDataExists
```
Angelmmiguel: I'm new on `Go`, so sorry about the question :P. Why do you capitalize the first letter of variables? Is this a convention?

src/api/data/cache/charthelper/icon_helper.go
```diff
@@ -0,0 +1,142 @@
...
+const originalFormat = "original"
+
+type iconFormat struct {
+	name   string
+	width  int
+	height int
+}
+
+var availableFormats = []iconFormat{
+	{"100x100", 100, 100},
```
Angelmmiguel: Can we move these constants to a configuration file?

src/api/data/cache/charthelper/icon_helper.go
```diff
@@ -0,0 +1,142 @@
...
+
+const originalFormat = "original"
+
+type iconFormat struct {
+	name   string
+	width  int
+	height int
+}
+
+var availableFormats = []iconFormat{
```
Angelmmiguel: With the current configuration we are generating images that fill the given height and width. For example, for Traefik the output is:

lTraefik 100 x 100

The problem is the logo is cropped, so some parts of the logo are missing. For image backgrounds is okey because we want to fill it as much as possible. But, for cards we want to show the entire logo. I think we can add another attribute to `iconFormat` called `strategy` to switch between `fill` and `fit` methods:

```go
// Strategy resizing method type
type StrategyType uint8

const(
  FIT StrategyType = iota
  FILL
)

type iconFormat struct {
  name         string
  strategy     StrategyType
  width        int
  height       int
}


var availableFormats = []iconFormat{
  {"100x100", FILL, 100, 100},
  {"300x300", FILL, 300, 300},
  {"100x100fit", FIT, 100, 100},
  {"300x300fit", FIT, 300, 300},
}
```

This parameter allow us to use an `fill` or `fit` strategy to resize and crop the image. The result for the same logo is:

Traefik 100x100fit

I know this image doesn't measure 100 x 100px, but `fit` method ensures me that the maximum height or width will be 100px. So, I can center it using `background` properties in the CSS.

We need to update the `processIcon` method to check the strategy:

```go
var processed *image.NRGBA
if format.strategy == FILL {
  processed = imaging.Fill(orig, format.width, format.height, imaging.Center, imaging.Lanczos)
} else {
  processed = imaging.Fit(orig, format.width, format.height, imaging.Lanczos)
}

```

migmartri: @Angelmmiguel it makes a lot of sense, thanks a lot!

To reduce computation penalty I'd like to reduce the number of processed icons to the minimum. What's in your take the formats (width, height and strategy) that we will need?

src/api/data/cache/cache_test.go
migmartri: This means that it is exported so it can be referenced in other packages.

migmartri:
Thanks @jackfrancis for the review.

Yes, We should definitely iterate and decompose some of that code in different packages with common dependencies for code sharing.

migmartri:
@Angelmmiguel The icon changes have been merged. As you can see, now the UI does not show any icons since they are using the old `icon` property.

Now, as part of the chartVersion API we include the different icon sizes that have been processed:

For example, in http://localhost:8080/v1/charts/stable/jenkins  you get the following payload.

selection_043

PR #111, opened 2017-02-10
Include Analytics

### LINKED ISSUE #69: Add Google analytics support
It should have full support for the SPA actions, including search

Angelmmiguel:
I investigated how to include GA in Angular2 applications. First of all, I tried to implement a custom solution based on a `config.json` file that allow us to enable and disable the analytics. However, we couldn't use this because we cannot include `script` tags in angular templates. I must put the GA code in `index.html` file.

Finally, I used [angulartics/angulartics2](https://github.com/angulartics/angulartics2) project to send the `pageviews` events when the user changes the location. Also, this project has a good integration with events and components.

To prevent users to upload the GA keys, I put a placeholder for the identifier in `index.html`.

This closes #69

### DIFF
--- a/helm/monocular/templates/ui-deployment.yaml
+++ b/helm/monocular/templates/ui-deployment.yaml
@@ -32,9 +32,16 @@ spec:
         volumeMounts:
           - name: vhost
             mountPath: /bitnami/nginx/conf/vhosts
+        volumeMounts:
+          - name: config
+            mountPath: /app/assets/js
         resources:
 {{ toYaml .Values.ui.resources | indent 12 }}
       volumes:
       - name: vhost
         configMap:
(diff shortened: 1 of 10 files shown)


helm/monocular/templates/ui-deployment.yaml
```diff
@@ -32,9 +32,16 @@ spec:
...
+          - name: config
+            mountPath: /app/assets/js
         resources:
 {{ toYaml .Values.ui.resources | indent 12 }}
       volumes:
       - name: vhost
         configMap:
-          name: {{ template "fullname" . }}-ui
+          name: {{ template "fullname" . }}-ui-vhost
+      volumes:
```
migmartri: Two volumes keys?

Issue #44, opened 2016-10-19
SEO Friendly rendering


migmartri:
Implement a SEO friendly rendering of the current Angular SPA.

We can do this implementing server side rendering or any other cache mechanism.

Angelmmiguel:
Angular has a project for friendly rendering in PC called Angular Universal. It provides a CLI to initialize the project called `universal-cli`. This CLI is a fork of `angular-cli`, so I'm a bit afraid of the compatibility of our project and this library.

Btw, `angular-cli` is causing us several problems due webpack misconfiguration. Also, that cli is still in alpha version. Should we remove it?

Angelmmiguel:
I investigated more about Universal. The [angular/universal-starter](https://github.com/angular/universal-starter) project creates a basic Frontend application + ExpressJS backend server. This server pre-renders the code and Universal starting to switch between real and pre-rendered implementation. This is an [interesting article](https://medium.com/google-developer-experts/angular-universal-for-the-rest-of-us-922ca8bac84#.g7orn9ijm) about how Angular Universal works.

However, this project has a lot of boilerplate code. Also, we need to think about [AoT and JIT code](https://angular.io/docs/ts/latest/cookbook/aot-compiler.html) and carry with files like this:

```

/*
 * THIS IS TEMPORARY TO PATCH 2.1.1+ Core bugs
 */

/* tslint:disable */
let __compiler__ = require('@angular/compiler');
import { __platform_browser_private__ } from '@angular/platform-browser';
import { __core_private__ } from '@angular/core';
if (!__core_private__['ViewUtils']) {
    __core_private__['ViewUtils'] = __core_private__['view_utils'];
}



if (__compiler__ && __compiler__.SelectorMatcher && __compiler__.CssSelector) {
    (__compiler__).__compiler_private__ = {
        SelectorMatcher: __compiler__.SelectorMatcher,
        CssSelector: __compiler__.CssSelector
    }
}
```

Before adopt it, we need to think about the maturity of this project because it seems for me that this is a one way road. In addition, I didn't find any alternative for server rendering.

migmartri:
I have noticed that there are some rendering issues in the details page. Sometimes the server side page does not include the readme page. We might need to tweak the prerender configuration via the env vars settings.

Angelmmiguel:
Should we increase the timeout of `prerender`? I think it can be related with the response time of the README.

migmartri:
[...]

PR #113, opened 2017-02-11
Server side rendering

migmartri:
* Package prerender phantom backend in a container.
* Add it to our current Helm Chart.
* Configure UIs client to redirect crawlers.
* Added fragment notification for bots. 

Refs https://github.com/helm/monocular/issues/44


migmartri:
@Angelmmiguel, the loading issue was related to the requirement of having the int polyfills for the date pipe.

Issue #127, opened 2017-02-15
Increase number of replicas


migmartri:
So we get proper rolling release

PR #128, opened 2017-02-15
Bump UI K8s replicas

migmartri:


migmartri:
Closes https://github.com/helm/monocular/issues/127

PR #162, opened 2017-02-17
Add "Copy to clipboard" to installation commands

### LINKED ISSUE #157: Copy to clipboard in usage section
Improve the usage section by:

* Adding inputs for every command that will horizontally scroll the content
* Auto select the whole content on click and copy it to clipboard. We should show a message

Angelmmiguel:
I added "Copy To Clipboard" feature for the commands we display in the Chart Details view. I followed the approach that @migmartri defined in #157. 

I changed the background of the panel and added a border and box shadow. I want to highlight this section of the view.

Chart details

Also, when the user clicks on the button, we display an Snack Bar to confirm the action. I tried to use the `tooltip` component, but it's fired automatically on `hover` event, so we can't show it programatically.

Copied to clipboard confirmation

Finally, I added two GA events to the buttons.

This closes #157

### DIFF
--- a/src/ui/src/app/chart-details/chart-details-usage/chart-details-usage.component.html
+++ b/src/ui/src/app/chart-details/chart-details-usage/chart-details-usage.component.html
@@ -1,6 +1,6 @@
 <app-panel class="chart-details-usage" [border]=true>
   <div *ngIf=showRepoInstructions class="chart-details-usage__repository">
-    <h2 class="chart-details-usage__label">Add incubator repository</h2>
+    <h2 class="chart-details-usage__label">Add {{ chart.attributes.repo.name }} repository</h2>
     <md-input floatingPlaceholder=false [value]=repoAddInstructions></md-input>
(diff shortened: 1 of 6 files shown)


src/ui/src/app/chart-details/chart-details-usage/chart-details-usage.component.html
```diff
@@ -1,7 +1,20 @@
-<div class="chart-details-usage">
-  <label>Usage</label>
-  <code>
-    <p *ngIf="showRepoInstructions">{{ repoAddInstructions }}</p>
-    <p>helm install {{ cmdChartId }} --version {{ currentVersion }}</p>
-  </code>
-</div>
+<app-panel class="chart-details-usage" [border]=true>
+  <div *ngIf=showRepoInstructions class="chart-details-usage__repository">
+    <h2 class="chart-details-usage__label">Add incubator repository</h2>
```
migmartri: Please use chart.name instead of "incubator"

PR #163, opened 2017-02-17
Enable ng2-meta

Angelmmiguel:
To solve this issue, I forked the [ng2-meta repo](https://github.com/vinaygopinath/ng2-meta) and added two pull requests to solve the current issues:

* [AOT Support](https://github.com/vinaygopinath/ng2-meta/pull/30)
* [Fix for default overwriting already set tags for a page](https://github.com/vinaygopinath/ng2-meta/pull/19)

As you can see, the commits are present in my repo: https://github.com/Angelmmiguel/ng2-meta/commits/master.

However, this is a temporal solution. We need to think a better way to update the meta tags because it seems the library is abandoned.

### DIFF
--- a/src/ui/src/app/app.module.ts
+++ b/src/ui/src/app/app.module.ts
@@ -42,13 +42,10 @@ import { LoaderComponent } from './loader/loader.component';
 
 require('hammerjs');
 
-let config = new ConfigService()
 const metaConfig: MetaConfig = {
   //Append a title suffix such as a site name to all titles
   useTitleSuffix: true,
   defaults: {
-    title: config.appName,
-    titleSuffix: ` | ${config.appName}`,
     description: 'Discover & launch great Kubernetes-ready apps'
   }
 };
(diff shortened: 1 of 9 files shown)


Tuizi:
Thank you very much, I really need those 2 fix! @michelleN when does this PR will be merged? :)

src/ui/src/app/app.module.ts
```diff
@@ -41,16 +41,15 @@ import { LoaderComponent } from './loader/loader.component';
...
-//    title: 'Monocular',
-//    titleSuffix: ' | Monocular',
-//    description: 'Find the Helm chart you need or share your own Kubernetes apps'
-//  }
-//};
+const metaConfig: MetaConfig = {
+  //Append a title suffix such as a site name to all titles
+  useTitleSuffix: true,
+  defaults: {
+    title: 'Monocular',
```
migmartri: Can we use config.service appName value instead?

migmartri:
Just updated the PR to support `config.appName` in the metadata as well as show the repository name in `/charts/:repo/` route.

migmartri:
@Angelmmiguel I get this error when trying to build in AOT, any idea? 

````
ERROR in Error encountered resolving symbol values statically. Reference to a local (non-exported) symbol 'metaConfig'. Consider exporting the symbol (position 46:7 in the original .ts file), resolving symbol AppModule in /app/src/app/app.module.ts
```

migmartri:
In the end I needed to disable the suffix in order to use the appname in the title because it broke the --aot compilation.

We will re-visit it later.

PR #182, opened 2017-02-23
NGINX ingress controller chart instructions

migmartri:
NGINX ingress controller has been merged as stable Chart so some modifications in the deployment instructions were required.

cc/ @mgoodness

https://kubeapps.com/charts/stable/nginx-ingress

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -6,7 +6,7 @@ Monocular is packaged as a Helm Chart that can be found in the `/deployment/mono
 
 > **Prerequisites**
 >
-> The chart is configured and tested to be used alongside an NGINX Ingress controller. Please be sure that you have a running instance in your cluster before proceeding. You can install it by executing `helm install stable/nginx-ingress`.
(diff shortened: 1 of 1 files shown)


docs/deployment.md
```diff
@@ -6,9 +6,9 @@ Monocular is packaged as a Helm Chart that can be found in the `/deployment/mono
 
 > **Prerequisites**
 >
-> The chart is configured and tested to be used alongside an NGINX Ingress controller. Please be sure that you have a running instance in your cluster before proceeding. More information  [here](https://github.com/kubernetes/ingress/tree/master/controllers/nginx).
+> The chart is configured and tested to be used alongside an NGINX Ingress controller. Please be sure that you have a running instance in your cluster before proceeding. You can install it by executing `helm install stable/nginx-ingress`.
```
mgoodness: I recommend linking to the nginx-ingress README rather than giving the `helm install` command. Any user will likely need to provide custom values for it to install correctly into their cluster. Don't want to misrepresent what's really involved.

migmartri: It makes sense. Changed :)

Thanks for getting back to me

Miguel

mgoodness:
LGTM!

Issue #187, opened 2017-02-23
Improve importing process


migmartri:
The current importing/indexing process has the following deficiencies.

* The importation happens sequentially and this approach was fine until the moment we added extra processing during importation time (untar and extract README files and process icons), which made this process slow.
* While we are in the importation process, no request can access to the data being imported, which means that the **API stops serving requests**.

We should improve this process by for example:

During importation:

- [x] Do not lock at the API level, keeping serving the content of the index. 
- [x] Do not clean the cache when the importation starts. keep serving the previous copy of the index. Getting replaced once the successful import finishes.
- [x] Icons downloads and processing failures are acceptable for a chart to be valid. In which case we can send this process to the background using a go routine "ignoring" its output
- [x] We should also look into default timeouts https://github.com/helm/monocular/issues/138

Regarding the deployment. In order to avoid re-importation of all the charts. We should start thinking about mounting the cache directory in a volume.

migmartri:
@jackfrancis I'd love to get your take here.

jackfrancis:
@migmartri I'm not able to find any unnecessary read locks on the API, but the 2nd bullet point is addressed w/ #188

PR #188, opened 2017-02-23
fix(cache): replaces data all at once

jackfrancis:
This change replaces the in-memory cache of charts data only when a complete, updated collection of repo charts data has been remotely refreshed, instead of at the beginning of the data extraction/transformation process.

Fixes the most egregious symptoms listed in #187.

### DIFF
--- a/src/api/data/cache/cache.go
+++ b/src/api/data/cache/cache.go
@@ -146,7 +146,7 @@ func (c *cachedCharts) Refresh() error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


migmartri:
Thanks @jackfrancis 

So this fixes the second checkbox point [here](https://github.com/helm/monocular/issues/187) right? In any case, while we have the readLocks (first checkbox point) in the handlers we will still not be able to return this brand new previous index right?

Have you noticed any significant importation performance increase?

jackfrancis:
@migmartri This addressed the 2nd point in #187, yes. This does not address any optimization in the data importation process!

Issue #138, opened 2017-02-16
Add shorter timeout to image pulling during refresh


migmartri:
It seems that from time to time pulling for example the Jenkins logo timeouts blocking the whole update process for about 10 seconds:

We need to either:

* Reduce this timeout
* Make image pulling parallel

PR #190, opened 2017-02-24
Add timeout to chart indexer

migmartri:
Added 5 seconds client timeout instead of relying on the server defined one. This is just part of a bigger optimization effort refs by  https://github.com/helm/monocular/issues/187

Closes https://github.com/helm/monocular/issues/138


prydonius:
LGTM

Issue #194, opened 2017-02-27
Add Repo source URL  


migmartri:
This is some work required to fulfill this requirements https://github.com/helm/monocular/issues/181 

Currently, repositories are defined using the following struct:

```go
var official = Repos{
	Repo{
		Name: "stable",
		URL:  "http://storage.googleapis.com/kubernetes-charts",
	},
	Repo{
		Name: "incubator",
		URL:  "http://storage.googleapis.com/kubernetes-charts-incubator",
	},
}
```

As @jackfrancis pointed here, this struct does not include the Git repository used for generating the repo index.

We could do as he suggested 

- [x] Rename URL to registryURL
- [x] Add the extra field `sourceURL`, which should represent the Github repo path which joined with the chartName will be a valid chart source URL i.e https://github.com/kubernetes/charts/tree/master/incubator/consul
- [x] Expose this values via the API
- [x] Adapt the UI to support the new json property keyname
In our case:

* Stable: https://github.com/kubernetes/charts/tree/master/stable
* Incubator: https://github.com/kubernetes/charts/tree/master/incubator

PR #195, opened 2017-02-28
feat(repos) Add sourceURL

migmartri:
Closes https://github.com/helm/monocular/issues/194


migmartri:
@jackfrancis thanks for the note. 

I did it on purpose for the cases of repositories that do not have a public Git repo. Or a git repo that is not UI accessible. 

Making it optional, I want to make the UI aware of this possibility.

jackfrancis:
👍

Issue #181, opened 2017-02-23
Link to the source of the charts


grosser:
I don't like blindly installing things ... and contributing to the charts would be great too

for example https://kubeapps.com/charts/stable/gitlab-ce has a link to dockerhub ... and that's it :(

prydonius:
+1 linking to the chart source in https://github.com/kubernetes/charts would be great!

jackfrancis:
It would benefit the UX to differentiate between source code and documentation of the underlying project itself (which is what the current links data under "Sources" references) from the source code of the chart (which is what @grosser wants a convenient reference to).

This just a note that we may want to improve the clarity of the current "Sources" sub-heading if we implement a link to the chart source.

jackfrancis:
I don't think we're currently representing the data response w/ the information needed (source repo for chart dev). We're instead representing the registry URL, e.g.:

`"repo": {name: "stable", url: "http://storage.googleapis.com/kubernetes-charts"}`

If we want to add this add'l URL, we might want to be more explicit, for example replacing the `url` property above w/ `registryUrl`, and then adding a `sourceUrl` property to point to `https://github.com/kubernetes/charts` (for example).

PR #200, opened 2017-03-01
feat(ui) Source repository

migmartri:
Closes https://github.com/helm/monocular/issues/181

* Added source repository and home sections
* Renamed source as `related`

selection_152

### DIFF
--- a/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
+++ b/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
@@ -38,7 +38,7 @@ export class ChartDetailsInfoComponent implements OnInit {
   get sourceName(): string {
     var parser = document.createElement('a');
     parser.href = this.chart.attributes.repo.sourceURL;
-    return parser.protocol + "//" + parser.hostname;
+    return parser.hostname;
   }
 
   loadVersions(chart: Chart): void {
--- a/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
+++ b/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
@@ -26,6 +26,21 @@ export class ChartDetailsInfoComponent implements OnInit {
     return this.chart.attributes.sources || [];
   }
 
+  get sourceUrl(): string {
+    var chartSource = this.chart.attributes.repo.sourceURL;
+    if (!chartSource) return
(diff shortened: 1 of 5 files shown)


src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
```diff
@@ -26,6 +26,21 @@ export class ChartDetailsInfoComponent implements OnInit {
...
+
+    // Used to handle possible trailing URLs
+    var urljoin = require('url-join');
+    return urljoin(chartSource, this.chart.attributes.name);
+  }
+
+  get sourceName(): string {
+    var parser = document.createElement('a');
+    parser.href = this.chart.attributes.repo.sourceURL;
+    return parser.protocol + "//" + parser.hostname;
```
prydonius: IMO the protocol should be removed for the sourceName

src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
```diff
@@ -26,6 +26,21 @@ export class ChartDetailsInfoComponent implements OnInit {
     return this.chart.attributes.sources || [];
   }
 
+  get sourceUrl(): string {
+    var chartSource = this.chart.attributes.repo.sourceURL;
+    if (!chartSource) return
+
+    // Used to handle possible trailing URLs
+    var urljoin = require('url-join');
```
prydonius: shouldn't requires be at the top? (not familiar with the way we're doing this though)

src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
migmartri: OK, removed. I added it since the rest of the links in related appear with the protocol. But since now that I have changed the order (being on top of maintainers), removing the protocol makes sense.

src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
migmartri: I just followed the official documentation.

https://github.com/jfromaniello/url-join#usage

Angelmmiguel: Yes, I think it should be at the top of the file

Issue #201, opened 2017-03-02
Expose available repositories via API


migmartri:
In order to implement https://github.com/helm/monocular/issues/198 we need the API to expose the information about the available repositories

PR #203, opened 2017-03-02
Enabled repositories API endpoint

migmartri:
* /v1/repos API route
* Using `repos` nickname based on Github's API schema

selection_154

Closes https://github.com/helm/monocular/issues/201


migmartri:
Regarding the message "No report found to compare against". I honestly don't know what I've done to break the coverage support.

@jackfrancis any idea what could be wrong?

Angelmmiguel:
@migmartri It seems `Codecov` doesn't find coverage result for the base commit. Can it be related with the S3 outage on Tuesday? 

More info: https://docs.codecov.io/docs/error-reference#section-missing-base-commit.

Issue #198, opened 2017-03-01
Adapt UI to support 3rd party repositories


migmartri:
Currently, the chart section is hardcoded to incubator and stable.

Related to https://github.com/helm/monocular/issues/191

migmartri:
Blocked by https://github.com/helm/monocular/issues/201

PR #206, opened 2017-03-02
3rd party repositories UI changes

migmartri:
Closes https://github.com/helm/monocular/issues/198

* Removes official charts UI coupling
* Uses new /v1/repos API endpoint

Example:

Homepage:

selection_156

Charts filtering page:

selection_157

Details page:
selection_158


src/ui/src/app/charts/charts.component.ts
```diff
@@ -30,27 +34,38 @@ export class ChartsComponent implements OnInit {
   }
 
   ngOnInit() {
+		this.loadRepos();
```
prydonius: weird indenting?

PR #213, opened 2017-03-06
Concurrent updates

migmartri:
Refs https://github.com/helm/monocular/issues/187

This PR aims to improve the importation times by:

* Use a buffered channel allowing a max of 25 concurrent downloads (the number is based on some empirical tests, imports started failing over 75, but 25+ workers made minor import time improvements) and assets processing.
* Use exclusive locking only at the moment of the cache update (Import does not seem to depend in any existing data). Allowing the API to keep accepting requests during importation.
* During import, store the new indexed data in a new map, transferring it as a whole if everything went well at the end of the importation.
* Moved initial refresh() foreground since now it does not depend on the Mutex
* Modified periodic package to support first run flag. Disabled by default once the app bootstraps.

### DIFF
--- a/src/api/data/cache/auto_refresher.go
+++ b/src/api/data/cache/auto_refresher.go
@@ -42,7 +42,7 @@ func (r *refreshChartsData) Frequency() time.Duration {
 	return r.frequency
 }
 
-// Frequency implements the Periodic interface
+// FirstRun implements the Periodic interface
 func (r *refreshChartsData) FirstRun() bool {
 	return r.firstRun
 }
--- a/src/api/data/cache/auto_refresher.go
+++ b/src/api/data/cache/auto_refresher.go
@@ -11,6 +11,7 @@ type refreshChartsData struct {
 	chartsImplementation data.Charts
 	frequency            time.Duration
 	name                 string
+	firstRun             bool
 }
 
 // NewRefreshChartsData creates a new Periodic implementation that refreshes charts data.
@@ -18,11 +19,13 @@ func NewRefreshChartsData(
 	chartsImplementation data.Charts,
 	frequency time.Duration,
 	name string,
+	firstRun bool,
 ) jobs.Periodic {
 	return &refreshChartsData{
 		chartsImplementation: chartsImplementation,
 		frequency:            frequency,
 		name:                 name,
+		firstRun:             firstRun,
 	}
 }
 
@@ -37,6 +40,11 @@ func (r *refreshChartsData) Do() error {
 // Frequency implements the Periodic interface
 func (r *refreshChartsData) Frequency() time.Duration {
 	return r.frequency
+}
+
+// Frequency implements the Periodic interface
+func (r *refreshChartsData) FirstRun() bool {
+	return r.firstRun
 }
 
 // Name implements the Periodic interface
--- a/src/api/data/cache/cache.go
+++ b/src/api/data/cache/cache.go
@@ -127,18 +127,18 @@ func (c *cachedCharts) Search(params charts.SearchChartsParams) ([]*models.Chart
 // Refresh is the interface implementation for data.Charts
 // It refreshes cached data for all authoritative repository+chart data
 func (c *cachedCharts) Refresh() error {
-	c.rwm.Lock()
-	defer c.rwm.Unlock()
+	// New list of charts that will replace cached charts
+	var updatedCharts = make(map[string][]*models.ChartPackage)
 
 	log.WithFields(log.Fields{
 		"path": charthelper.DataDirBase(),
 	}).Info("Using cache directory")
 
 	for _, repo := range c.knownRepos {
-		// Append index.yaml
 		u, _ := url.Parse(repo.URL)
 		u.Path = path.Join(u.Path, "index.yaml")
 
+		// 1 - Download repo index
 		resp, err := http.Get(u.String())
 		if err != nil {
 			return err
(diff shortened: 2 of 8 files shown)


src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
+		// 2 - Parse repo index
 		charts, err := helpers.ParseYAMLRepo(body, repo.Name)
 		if err != nil {
 			return err
 		}
+
+		// 3 - Process elements in index
 		var chartsWithData []*models.ChartPackage
+		// Buffered channel
+		ch := make(chan chanItem, len(charts))
```
migmartri: Using buffered channels to avoid deadlocks and routine leak in case of error in the drain procedure.

src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
-					// Skip chart if error extracting the tarball
-					continue
-				}
-				// If we have a problem processing an image it will fallback to the default one
-				charthelper.DownloadAndProcessChartIcon(chart)
+			go processChartMetadata(chart, ch)
+		}
+		// 3.2 Channel drain
+		for range charts {
+			it := <-ch
```
migmartri: Channel drain step.

src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
 			}
-			chartsWithData = append(chartsWithData, chart)
 		}
-		c.allCharts[repo.Name] = chartsWithData
+		updatedCharts[repo.Name] = chartsWithData
 	}
+
+	// 4 - Update the stored cache with the new elements if everything went well
+	c.rwm.Lock()
+	c.allCharts = updatedCharts
```
migmartri: Full cache replacement.

src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
+type chanItem struct {
+	chart *models.ChartPackage
+	err   error
+}
+
+// Counting semaphore, 25 downloads max in paralell
+var tokens = make(chan struct{}, 25)
+
+func processChartMetadata(chart *models.ChartPackage, out chan chanItem) {
+	tokens <- struct{}{}
```
migmartri: Buffered channel following the counting semaphore pattern.

src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
+		<-tokens
+	}()
+
+	var it chanItem
+	it.chart = chart
+
+	// Extra files. Skipped if the directory exists
+	dataExists, err := charthelper.ChartDataExists(chart)
+	if err != nil {
+		it.err = err
```
migmartri: Error handling done via sending errors to the channel via the chanItem struct

src/api/swagger/restapi/configure_monocular.go
```diff
@@ -37,10 +37,14 @@ func configureAPI(api *operations.MonocularAPI) http.Handler {
 	}
 	// configure the api here
 	chartsImplementation := cache.NewCachedCharts(config.Repos)
+	// Run foreground repository refresh
+	chartsImplementation.Refresh()
```
migmartri: Move refresh to foreground during app bootstrap because technically before we were blocking anyways but via a `Sync.Lock` which took into effect once called doPeriod(). This method was brittle and prone to race conditions.

src/api/data/cache/auto_refresher.go
```diff
@@ -39,6 +42,11 @@ func (r *refreshChartsData) Frequency() time.Duration {
 	return r.frequency
 }
 
+// Frequency implements the Periodic interface
```
jackfrancis: change "Frequency" to "FirstRun" in the method declaration comment

src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
+// Represents every element processed in paralell
+type chanItem struct {
+	chart *models.ChartPackage
+	err   error
+}
+
+// Counting semaphore, 25 downloads max in paralell
+var tokens = make(chan struct{}, 25)
+
+func processChartMetadata(chart *models.ChartPackage, out chan chanItem) {
```
jackfrancis: Because out is only used as a send channel, you can change the 2nd parameter specification to `out chan<- chanItem`.

src/api/data/cache/cache.go
```diff
@@ -127,18 +127,18 @@ func (c *cachedCharts) Search(params operations.SearchChartsParams) ([]*models.C
 // Refresh is the interface implementation for data.Charts
 // It refreshes cached data for all authoritative repository+chart data
 func (c *cachedCharts) Refresh() error {
-	c.rwm.Lock()
```
jackfrancis: Are you sure we want to remove mutex locking here? Do we want to enable concurrent `Refresh()` executions?

technosophos: The only issue I see with concurrency here is that because all refresh calls share the same pool of workers, it would be possible to negatively impact the performance of the refresh merely by calling the function a whole bunch of times.

src/api/data/cache/cache.go
technosophos: I can't find where `ch` is closed. You might be leaking channels here.

src/api/data/cache/cache.go
```diff
@@ -148,37 +148,84 @@ func (c *cachedCharts) Refresh() error {
...
-			chartsWithData = append(chartsWithData, chart)
 		}
-		c.allCharts[repo.Name] = chartsWithData
+		updatedCharts[repo.Name] = chartsWithData
 	}
+
+	// 4 - Update the stored cache with the new elements if everything went well
+	c.rwm.Lock()
+	c.allCharts = updatedCharts
+	defer c.rwm.Unlock()
```
technosophos: You can remove `defer` here, since there is no event that could trigger between this and `return`.

src/api/data/cache/cache.go
migmartri: I thought that closing channels were only needed to be done from the sender to indicate the receiver from this fact.

According to this post (I also read it in A.A Donovan's book), leaving a channel open without senders will be automatically garbage collected. http://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open https://groups.google.com/forum/#!msg/golang-nuts/pZwdYRGxCIk/qpbHxRRPJdUJ

It am ok closing it, I want to know if in your opinion that's a best practice.

src/api/data/cache/cache.go
migmartri: I am not trying to optimize for running `Refresh()` multiple times, actually in theory there should be only one goRoutine being executed at once (unless it takes longer than the refresh period).

What I want to allow is that during the refresh process we keep accepting API calls. 

With the current implementation, assuming that for example the full refresh took 20 seconds (it downloads and process stuff). That would mean that no API requests will be accepted because of the Lock in Refresh and Rlocks in the auxiliare cache methods during all that time. https://github.com/helm/monocular/blob/master/src/api/data/cache/cache.go#L85

With the new implementation:

Downloads of new charts and assets happen without blocking the structure, just locking it at the moment of the content swap, having virtually no downtime.

> The only issue I see with concurrency here is that because all refresh calls share the same pool of workers, it would be possible to negatively impact the performance of the refresh merely by calling the function a whole bunch of times.

Could you elaborate? Is this related to tweaking the number of workers? Do you have any suggestions on how to measure and tweak it?

src/api/data/cache/cache.go
migmartri: Oops. Thanks

jackfrancis:
@migmartri LGTM, ready for merge!

Issue #193, opened 2017-02-27
In Cluster Helm integration


migmartri:
Currently, Monocular shows the instructions on how to install a selected chart using the CLI.

![selection_147](https://cloud.githubusercontent.com/assets/24523/23381506/92684fe6-fcf3-11e6-9e91-888f87af3b03.png)

What if we could offer a tighter integration with Helm that will make Monocular show an "Install" button which will trigger an in-cluster installation of the chart?

This integration will require:

> **NOTE: Required features for a first version marked as MVP**

### Launch capabilities

Once clicked the install button:  

* Initialize tiller in the cluster if needed. 
* Show a form with the values that can be configured (read from values.yaml).
* MVP: Execute a `helm install` or similar approach (talk with Tiller directly, use Helm's engine.go package, etc). 

### Releases list

* Monocular could include a releases list of charts running in the cluster. This could be as simple as proxying `helm list` via the backend JSON API.
* In that list, there must be a delete button that will uninstall that release. 

### Configuration options

* MVP: Any instance of Monocular running in an K8s cluster should be able to enable this feature via en environment variable/config map.

#### Deploy a Chart Version `POST /releases`
 
Arguments:

```
chartName: string (MVP)
chartVersion: string (MVP)
namespace: string
releaseName: string
values: object
```

Response: 201, data: releaseObject.
 
#### Get the list of releases `GET /releases`

Response: 200, data: releaseObject[]

#### Delete a release `DELETE /releases/{RELEASE_NAME}`

Response: 200

### Release object

```
{
  id: <ReleaseName>
  type: release
  attributes: {
    updated: <updatedAt>,
    chart: <chart>,
    status: <status>
}

Thoughts?

migmartri:
@mattfarina, I'd love to get your thoughts on this. Is this similar to what you had in mind?

Last year, I opened this issue regarding a similar integration in the K8s dashboard https://github.com/kubernetes/helm/issues/990

But I think that doing the integration first in Monocular will be useful to test if traction and usefulness before proposing adding it upstream in the dashboard. In any case, a lot of the work done here could be re-used there (and vice versa) 

cc/ @technosophos @jackfrancis @Angelmmiguel @prydonius

bcorijn:
Learned about this after joining the Helm standup yesterday, definitely sounds as a nice feature. I certainly see the use in my organisation for allowing easier deploys by developers that are not necessarily very knowledgeable about Helm or Kubernetes. 

A few features that would be great for our use-case:
[...]

PR #221, opened 2017-03-15
Improve the UI of Deployments

Angelmmiguel:
I improved the design of the `/deployments` views. Mayor improvements are:

* Rename all release components to deployments
* Show the deployments as a list in `/deployments`
* Update the layout of Deployment view
* Parse Resources and display the information as a table
* Improve modals
* Add new icons

Refs #193 

#### Deploy 
Deploy
#### Deployment list 
Deployment list
#### Deployment list hover effect
Deployment list hover effect
#### Deployment details
Deployment details
#### Delete a deployment
Deployment remove


migmartri:
Nice!

* Where does the "new deployment" button points to?
* Any reason why we are using different color buttons in the delete and in the deployment confirmation modals?

Angelmmiguel:
> Where does the "new deployment" button points to?

This button redirects the user to `/charts`. From this route he will be able to select an application and deploy it.

> Any reason why we are using different color buttons in the delete and in the deployment confirmation modals?

I think it's important to highlight the operation the user is going to perform. Deleting a resource is an important operation, so I think it's better to use the `warning` color. Launching a new deploying is not a "critical" action, it can be deleted easily. That's the reason I used `primary` color for this dialog.

PR #228, opened 2017-03-23
SEO improvements

Angelmmiguel:
Currently, every component set the titles and descriptions of the pages. This means, these values are distributed over the project, so we don't have a common point with this information.

To solve this, I've create a new service that configures the metatags for a given page. It allows you to use placeholders that will be overridden by the provided information from the component. The titles and description are located in `src/ui/src/app/shared/seo.data.ts`. As you can see, there are several placeholders in the code (Ex. `{ repo }`).

To update the metatags, you only need to call the `setMetaTags(page, values)` method of the service.

### DIFF
--- a/src/ui/src/app/shared/seo.data.ts
+++ b/src/ui/src/app/shared/seo.data.ts
@@ -9,17 +9,17 @@ export default {
   },
   search: {
     title: 'Results for "{ search }" Kubernetes-ready applications',
-    description: 'Results for "{ search }" in the KubeApps catalog of Kubernetes-ready applications. ' +
+    description: 'Results for "{ search }" in the { appName } catalog of Kubernetes-ready applications. ' +
                  'Deploy all apps you need in your infrastructure or the cloud'
   },
   charts: {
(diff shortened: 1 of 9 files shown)


src/ui/src/app/shared/seo.data.ts
```diff
@@ -0,0 +1,34 @@
...
+                 'apps. Browse the catalog and deploy your applications in your Kubernetes cluster'
+  },
+  search: {
+    title: 'Results for "{ search }" Kubernetes-ready applications',
+    description: 'Results for "{ search }" in the KubeApps catalog of Kubernetes-ready applications. ' +
+                 'Deploy all apps you need in your infrastructure or the cloud'
+  },
+  charts: {
+    title: 'Kubernetes-ready applications catalog',
+    description: 'Browse the KubeApps catalog of Kubernetes-ready apps. Deploy all apps you need ' +
```
migmartri: Kubeapps can not be hardcoded

Angelmmiguel:
@migmartri Thank you for notice about it. I've already fixed the descriptions.

Issue #230, opened 2017-04-05
Search in monocular does not work


rfranzke:
When searching for a chart in Monocular, nothing happens (basically a never ending search icon is shown).
This issue seems to apply to Monocular in general as it does neither work for the deployment in my own K8S cluster nor at kubeapps.com

prydonius:
Hmm, this must be due to a recent change because search was working last week.

I'm getting the following error in the console:

```
TypeError: Cannot read property 'map' of null
    at https://kubeapps.com/main.4f8de63d4069147584bf.bundle.js:5:15713
    at Array.filter (native)
    at e.project (https://kubeapps.com/main.4f8de63d4069147584bf.bundle.js:5:15515)
    at e._next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:317:27302)
    at e.next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:8:3990)
    at e._next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:8:4319)
    at e.next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:8:3990)
    at e._next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:1445:1828)
    at e.next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:8:3990)
    at e._next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:317:27400)
    at e.next (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:8:3990)
    at XMLHttpRequest.f (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:611:666)
    at t.invokeTask (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:1482:10025)
    at Object.onInvokeTask (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:141:2559)
    at t.invokeTask (https://kubeapps.com/vendor.051d5f300f158e6a8ad6.bundle.js:1482:9946)
```

Angelmmiguel:
Hello @rfranzke,

Thank you for reporting the issue. It seems some of the charts doesn't include the maintainers array in the metadata. I'm going to check if it's present before try to search in the array.

Angelmmiguel:
[...]

PR #233, opened 2017-04-05
Several fixes for the new version of angular packages

Angelmmiguel:
There are some errors with the new versions of the packages on `--aot` compilation. This PR fix all of them.

Refs #230


Angelmmiguel:
There was another error caused by `angular/material`. I'm checking it.

PR #239, opened 2017-04-21
Bump health check for initial delay for monocular-api

nomisbeme:
I'm running monocular under minikube (under xhyve) on a less than powerful 8GB MacBook Air.

After deploying tiller & nginx-ingress I was able to deploy monocular using:
helm install deployment/monocular

On my system the monocular-api pod gets killed before it has started up by the default health check determining the pod is unhealthy, resulting in a slow motion crash-loop. 

From looking at the logs, the monocular-api pod is not unhealthy it's just doing the first (synchronous) poll of the stable repo which takes longer than the initial health check timeout (60 seconds.)

This PR hacks around the problem by upping the initial health check timeout. A proper fix would probably make the initial poll asynchronous or implement a more sophisticated set of health checks.

### DIFF
--- a/deployment/monocular/templates/api-deployment.yaml
+++ b/deployment/monocular/templates/api-deployment.yaml
@@ -23,13 +23,13 @@ spec:
           httpGet:
             path: /healthz
             port: {{ .Values.api.service.internalPort }}
-          initialDelaySeconds: 60
+          initialDelaySeconds: 180
           timeoutSeconds: 10
         readinessProbe:
           httpGet:
             path: /healthz
             port: {{ .Values.api.service.internalPort }}
-          initialDelaySeconds: 30
+          initialDelaySeconds: 90
           timeoutSeconds: 5
         volumeMounts:
           - name: config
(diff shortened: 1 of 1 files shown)


deployment/monocular/templates/api-deployment.yaml
```diff
@@ -23,13 +23,13 @@ spec:
...
             port: {{ .Values.api.service.internalPort }}
-          initialDelaySeconds: 60
+          initialDelaySeconds: 180
           timeoutSeconds: 10
         readinessProbe:
           httpGet:
             path: /healthz
             port: {{ .Values.api.service.internalPort }}
-          initialDelaySeconds: 30
+          initialDelaySeconds: 90
```
prydonius: I think we can leave this as-is as the readiness probe will just determine whether the pod is reachable, but won't kill it.

nomisbeme: New PR here https://github.com/nomisbeme/monocular/pull/1
Or if you want this another way, let me know.

nomisbeme:
Sort for the confusion on my part. Hopefully this is a little better?

prydonius:
Thanks @nomisbeme!

PR #241, opened 2017-05-01
Provide clickable urls on deployment status page by parsing Service strings

tarasglek:
I really like monocular. However it's missing a major usability feature of being able to visit deployed services. This implements that.

I'm sorry I'm parsing a text string. I had to do this because there is no structured status data being passed from tiller. I view this as an intermediate solution until there is a yaml-output equivalent of helm status.

screen shot 2017-04-28 at 2 28 50 pm

### DIFF
--- a/src/ui/src/app/shared/services/deployments.service.ts
+++ b/src/ui/src/app/shared/services/deployments.service.ts
@@ -50,7 +50,99 @@ export class DeploymentsService {
 
   private extractData(res: Response, fallback = {}) {
     let body = res.json();
-    return body.data || fallback;
+    var data = body.data;
+    if (!data) {
+        return fallback;
+    }
+    var attributes = data.attributes;
+    if (attributes) {
+      attributes.urls = [];
+      var resources = attributes.resources;
+      if (resources) {
+        var parsedResources = this.loadResources(data);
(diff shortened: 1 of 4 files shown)


src/ui/src/app/shared/services/deployments.service.ts
```diff
@@ -50,7 +50,79 @@ export class DeploymentsService {
...
+      }
+    })
+  }
+
+  /**
+   * Parse resource string and return sections as tables
+   * @param {string} resourceStr 'helm status' output
+   * @returns object mapping section to array of contents 
+   */
+  private parseResources(resourceStr) {
```
migmartri: Any chance we could use/extend/replace the parser that is part of https://github.com/helm/monocular/blob/master/src/ui/src/app/deployment/deployment.component.ts#L57 ?

tarasglek:
Updated to use pre-existing parser. Thanks for pointing it out.

tarasglek:
@Angelmmiguel, incorporated changes addressing your feedback. Can we merge this now?

Angelmmiguel:
Yes! Thank you :)

migmartri:
Thanks @tarasglek and @Angelmmiguel!

PR #261, opened 2017-05-23
Add OWNERS file

prydonius:
cc @migmartri @Angelmmiguel @jackfrancis @technosophos 

@technosophos I've put you down as an assignee since you've helped review PRs, but let me know if you're prefer not be listed here.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,6 @@
+assignees:
+- migmartri
+- Angelmmiguel
+- jackfrancis
+- technosophos
+- prydonius
--- a/OWNERS
+++ b/OWNERS
@@ -1,6 +1,6 @@
 assignees:
-- migmartri
-- Angelmmiguel
-- jackfrancis
-- technosophos
-- prydonius
+  - migmartri
+  - Angelmmiguel
+  - jackfrancis
+  - technosophos
+  - prydonius


OWNERS
```diff
@@ -0,0 +1,6 @@
+assignees:
```
Angelmmiguel: What's the difference between `assignees` and `maintainers`?

OWNERS
```diff
@@ -0,0 +1,6 @@
...
+- migmartri
```
Angelmmiguel: In the Kubernetes repo, the [OWNERS file](https://github.com/kubernetes/charts/blob/master/OWNERS) has a different indentation. I'm not sure if this kind of files are standardized

OWNERS
prydonius: This seems to be inconsistent everywhere (core uses approvers and reviewers)... I'm following the example in the template linked from the incubation doc: https://github.com/kubernetes/kubernetes-template-project/blob/master/OWNERS

OWNERS
prydonius: It's YAML so I believe it doesn't matter, but I'll change it to match the other repos :+1:

OWNERS
Angelmmiguel: :+1:

PR #264, opened 2017-06-01
add ROADMAP

prydonius:


prydonius:
cc @migmartri @Angelmmiguel @sebgoa

Angelmmiguel:
@prydonius should we map the items of this file to issues and create a board in projects? I think it more easy to update. We don't need to update this file every time we finish or update an issue

prydonius:
@Angelmmiguel yeah I was planning on setting up milestones/project after merging. I think we can see this as more of a higher level document, it doesn't necessarily need to link to issues. It might also make sense to have this in the wiki instead. I based this off of https://github.com/kubernetes-incubator/service-catalog/wiki/Roadmap

Angelmmiguel:
@prydonius I agree we can use the wiki. I think it's easier to update

prydonius:
@Angelmmiguel I've moved this to the wiki https://github.com/helm/monocular/wiki/Roadmap

thanks!

Angelmmiguel:
Thanks @prydonius!!

PR #266, opened 2017-06-06
Enable configuring CORS to remove same-domain requirement

### LINKED ISSUE #235: Support Helm deployment without Ingress controller
Currently, Monocular requires the use of an Ingress Controller by having `/api` as default. This was done in order to simplify the setup and avoid CORS issues.

We want to support deployment in which backend and frontend are in different hostnames.

* Configure the UI to use this hostname for communications with the backend. This is currently supported (but not used) by passing the `backendHostname` property in the overrides.js. This file gets mapped using the config map `ui-config.yaml`. We need to extend this config map to read a new value in the values.yaml.

* In the backend, whitelist the UI domain using the CORS configuration module. We also need to We add a new option in the values.yaml to define the UI hostname. For example `frontendHostname`.

CORS is currently configured in `src/api/config/cors/cors.go`. Currently it only whitelist everything if `ENVIRONMENT=development`. We should extend this to read the domain `frontendHostname` via env variables or command line flags.

Summarizing:

1 - Add `frontendHostname` and `backendHostname` properties in the values.yaml file.
2 - Change  `ui-config.yaml` to expose `backendHostname`
3 - Modify `src/api/config/cors/cors.go` to whitelist `frontendHostname`

cc/ @ursuad @prydonius  refs #234

prydonius:
fixes #235

### DIFF
--- a/deployment/monocular/templates/ui-config.yaml
+++ b/deployment/monocular/templates/ui-config.yaml
@@ -13,6 +13,9 @@ data:
       overrides: {
         googleAnalyticsId: '{{.Values.ui.googleAnalyticsId}}',
         appName: '{{.Values.ui.appName}}',
+        {{- if .Values.ui.backendHostname }}
+        backendHostname: '{{ .Values.ui.backendHostname }}',
+        {{- end }}
         releasesEnabled: {{.Values.api.config.releasesEnabled}}
       }
     };
--- a/deployment/monocular/values.yaml
+++ b/deployment/monocular/values.yaml
@@ -44,7 +44,6 @@ api:
     releasesEnabled: false
     # Cache refresh interval in sec.
     cacheRefreshInterval: 3600
-
 ui:
   replicaCount: 2
   image:
@@ -66,6 +65,7 @@ ui:
   # ui-config populate
   googleAnalyticsId: UA-XXXXXX-X
   appName: Monocular
+  # backendHostname: http://monocular-api.local
 
 prerender:
   replicaCount: 1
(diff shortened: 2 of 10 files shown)


deployment/monocular/templates/ui-config.yaml
```diff
@@ -13,6 +13,9 @@ data:
       overrides: {
         googleAnalyticsId: '{{.Values.ui.googleAnalyticsId}}',
         appName: '{{.Values.ui.appName}}',
+        {{- if .Values.ui.backendHostname }}
+        backendHostname: '{{ .Values.ui.backendHostname }}',
```
prydonius: I decided not to make the default `/api` to not change the way it currently work, but maybe that makes sense?

Angelmmiguel: I think it makes sense to remove it because we can use a subdomain like `api.mydomain.com`. Adding `/api` may be repetitive. However, I think this is not so important now and it's a different issue.

deployment/monocular/values.yaml
```diff
@@ -34,11 +34,16 @@ api:
       #- name: my-repo-name
       #  url: my-repository-url
       #  source: my-repository-source
+    cors:
+      allowed_origins:
+        - my-api-server
```
Angelmmiguel: I see this is a configurable parameter. Can you add some documentation? I'm not sure if the frontend will work with this default configuration.

prydonius: This is the default in https://github.com/helm/monocular/blob/master/src/api/config/cors/cors.go#L30, I just copied it here. It means that only a hostname of `my-api-server` or same-origin will work. A blank value would allow all origins, I believe.

Angelmmiguel: You're right :+1:. Thanks for the explanation

PR #267, opened 2017-06-07
Improve tiller communication to support RBAC

prydonius:
- removes querying the K8S API to check if tiller is installed
- removes attempt to install tiller (this should be handled by the
  cluster admin as it requires setting up service accounts and roles in
  RBAC environments)
- only accesses the K8S API if tillerPortForward setting is enabled

### DIFF
--- a/src/api/data/helm/client/client.go
+++ b/src/api/data/helm/client/client.go
@@ -59,7 +59,7 @@ func (c *helmClient) initialize() (*helm.Client, error) {
 	client := helm.NewClient(helm.Host(tillerHost))
 	// test connection
 	if _, err := client.GetVersion(); err != nil {
-		return nil, errors.New("Failed to connect to Tiller, are you sure it is installed?")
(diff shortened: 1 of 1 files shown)


src/api/data/helm/client/client.go
```diff
@@ -80,7 +56,13 @@ func (c *helmClient) initialize() (*helm.Client, error) {
 		tillerHost = fmt.Sprintf("localhost:%d", tunnel.Local)
 	}
 
-	return helm.NewClient(helm.Host(tillerHost)), nil
+	client := helm.NewClient(helm.Host(tillerHost))
+	// test connection
+	if _, err := client.GetVersion(); err != nil {
+		return nil, errors.New("Failed to connect to Tiller, are you sure it is installed?")
```
mmikulicic: error strings should not be capitalized. Read more about it in https://github.com/golang/go/wiki/CodeReviewComments#error-strings

prydonius:
thank you for the review @mmikulicic!! :)

PR #275, opened 2017-06-15
Fixes parsing of service URL

### LINKED ISSUE #246: Generated URL Link is incorrect
Steps:
1. Launch a chart using the Install button
2. Wait for the application to install and be ready
3. Find the URL link on the right hand side and click it

Result:
A browser window opens, but the application never loads.

Expected:
The application loads

Notes:
Removing the added port number form the URL results in a working URL.
I found this on dokuwiki and mediawiki.

prydonius:
The original expression was parsing had INT_PORT:EXT_PORT/TCP, but it
seems this has changed to EXT_PORT:INT_PORT/TCP.

Fixes #246

### DIFF
--- a/docs/config.example.yaml
+++ b/docs/config.example.yaml
@@ -13,12 +13,12 @@ repos:
   #  url: my-repository-url
   #  source: my-repository-source
 
-cors:
-  allowed_origins:
-    - my-api-server
-  allowed_headers:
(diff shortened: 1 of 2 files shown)


prydonius:
screen shot 2017-06-15 at 21 22 05

PR #279, opened 2017-06-20
fix chart default api host

### LINKED ISSUE #278: Overriding releasesEnabled disables API_NAME override
In issue #277, I was attempting to override some of the default settings from the CLI.   The default setting today for Monocular is to disable releasesEnabled.   I was able to enable deployments by setting the values.yaml property to true but the command line appears to override my API_NAME setting for some reason (getting CORS errors when I add in VALUES_OVERRIDE thus appearing to ignore my API_NAME).    This might be a user error but wanted to document this in case others hit the same issue.  

API_NAME="https://$ipAddress:$port" API_TAG=latest UI_TAG=latest VALUES_OVERRIDE="api.config.releasesEnabled=true" ./deploy_install.sh

prydonius:
fixes #278

### DIFF
--- a/deployment/monocular/Chart.yaml
+++ b/deployment/monocular/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: monocular
 description: Monocular is a search and discovery front end for Helm Charts Repositories.
-version: 0.4.0
+version: 0.4.1
 appVersion: 0.2.0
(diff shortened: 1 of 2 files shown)


prydonius:
cc @todkap

prydonius:
CI is failing because of a small mismatch in the number of charts, I suspect that's because a new chart was recently added.

PR #281, opened 2017-06-21
clean chart cache dir on error processing

prydonius:


src/api/data/cache/charthelper/chart_package_helper.go
```diff
@@ -134,6 +136,19 @@ var ensureChartDataDir = func(chart *models.ChartPackage) error {
 	return nil
 }
 
+var cleanChartDataDir = func(chart *models.ChartPackage) error {
```
mmikulicic: why is this a variable? do you intend to replace it?

or do you do that for every function just in case?

src/api/data/cache/charthelper/chart_package_helper.go
```diff
@@ -134,6 +136,19 @@ var ensureChartDataDir = func(chart *models.ChartPackage) error {
...
+	dir := chartDataDir(chart)
+	if _, err := os.Stat(dir); err != nil {
```
mmikulicic: why do you want to return an error if the dir doesn't exist?

i.e. I wonder why cannot this just be:

```
var cleanChartDataDir = func(chart *models.ChartPackage) error {
  return os.RemoveAll(chartDataDir(chart))
}
```

src/api/data/cache/charthelper/chart_package_helper.go
```diff
@@ -29,11 +29,13 @@ var DownloadAndExtractChartTarball = func(chart *models.ChartPackage) error {
 
 	if !tarballExists(chart) {
 		if err := downloadTarball(chart); err != nil {
+			cleanChartDataDir(chart)
```
mmikulicic: Just a FYI:

```
var DownloadAndExtractChartTarball = func(chart *models.ChartPackage) (err error) {
 	if err := ensureChartDataDir(chart); err != nil {
 		return err
 	}

 	defer func() {
 	 	if err != nil {
 	 	 	cleanChartDataDir(chart)
 	 	}
 	}()

 	if !tarballExists(chart) {
  		if err := downloadTarball(chart); err != nil {
			return err
  		}
  	}

 	if err := extractFilesFromTarball(chart); err != nil {
  		return err
  	}
  
 	return nil
}
```

this ensures that if you add another place where you return an error you cannot forget to delete the char data dir when the error happens.

prydonius: Oh nice, I was trying to figure out how I could do this, will change it to this - thanks!

src/api/data/cache/charthelper/chart_package_helper.go
prydonius: This is something that we're doing for functions that need to be overridden in tests, although I'm not actually overriding it in this case. I'm not sure if it makes sense to only use a variable where we intend to test it that way. What do you think?

src/api/data/cache/charthelper/chart_package_helper.go
prydonius: Fair point, I was copying the same structure for `ensureDataDir` but I think you're right in that this could be much simpler.

Issue #231, opened 2017-04-05
Update monocular backend to respect /api/ path


dcberg:
Today the monocular ingress is using the ` ingress.kubernetes.io/rewrite-target:` annotation to rewrite `/api/` to `/`. This limits portability of monocular to only ingress controllers that support the rewrite annotation.

prydonius:
cc @migmartri - I think we talked about this before? Given that the frontend assumes the prefixed path, I don't think it would be a stretch for the backend to assume the prefixed path too. What do you think?

migmartri:
@dcberg good point, what ingress controller are you planning on running monocular with?

@prydonius yeah, I think that adding /api to backend's basepath seems reasonable.

dcberg:
Sorry for the delay @migmartri 

We have our own ingress controller that is based on nginx. Our current implementation doesn't support the rewrite annotation; however, we are adding the support.

mlaccetti:
We're using Traefik as the ingress controller, which doesn't seem to understand the `rewrite-target` annotation. It has a custom annotation (`traefik.frontend.rule.type: pathprefixstrip`), but it applies it to all paths, which mangles the UI paths.

prydonius:
@mlaccetti @dcberg I just added support for specifying separate hostnames for the API and frontend via the [`ui.backendHostname`](https://github.com/helm/monocular/blob/master/deployment/monocular/values.yaml#L68) property which configures the API location and the [`api.config.cors.allowed_origins`](https://github.com/helm/monocular/blob/master/deployment/monocular/values.yaml#L37) property which configures CORS to accept requests from your UI hostname.

With these you can avoid using a `/api` path and instead use subdomains such as `api.monocular.local` and `monocular.local`:

```
helm install ./deployment/monocular --set ui.backendHostname=https://api.monocular.local,api.config.cors.allowed_origins={http://monocular.local}
```

mlaccetti:
This will require changing the ingress to add two different domains, then?

prydonius:
@mlaccetti yes, I haven't changed the ingress template in the chart for this, but I was thinking of adding an `ingress.enabled` flag which can be used to use your own ingress. I will add this in combination with updating the template to use the hostnames if they are set.

prydonius:
@mlaccetti I have added an `ingress.enabled` flag to the chart. If you need to serve Monocular on separate domains, at the moment, you will need to setup your own Ingress resource independently of the chart.

PR #286, opened 2017-06-22
Update chart and documentation

### LINKED ISSUE #73: Be able to configure backendURL via env variables
Currently, the UI has the backend domain and port [hardcoded](https://github.com/helm/monocular/blob/master/src/ui/src/app/config.ts#L6). We must find a way to set those values via env variables. 

In theory this is feasible using webpack but last time I looked into it I could not find a simple solution since angular-ui bundles its own webpack configuration and did not allow extending it.

### LINKED ISSUE #272: Serve from a chart repository
- Add to kubernetes/charts stable
- Add a github pages repository to this repo to serve latest releases

prydonius:
Updated the documentation to reference the chart repository (fixes #272).

Also improved configuration documentation and explained how to setup monocular to serve on separate domains. Also added an option to disable ingress in this case (fixes #73 #231)


prydonius:
cc @sameersbn @migmartri

sameersbn:
LGTM!

PR #293, opened 2017-06-26
Support appVersion field

prydonius:
Adds support for the appVersion field from chart metadata.

The appVersion will be shown instead of the chart version in a few prominent places (e.g. chart list view), as this is more useful than the chart version.

screen shot 2017-06-27 at 14 09 36

screen shot 2017-06-27 at 16 46 54

This PR also moves to use the github.com/ghodss/yaml library instead of gopkg.in/yaml.v2. This library works better with the Swagger models as it uses the `json` tags on structs which Swagger generates.

On the UI side, I modified the chart-details component to fetch the ChartVersion object rather than just pass a string to the children. This made it possible to add the "Application Version" section to the details, and is better as it prevents the child components from having to fetch the version if they need it (e.g. in chart-details-readme).


prydonius:
cc @Angelmmiguel @kemcake

kemcake:
Tested 👍 lgtm

PR #297, opened 2017-06-27
Move all ingress annotations to values

### LINKED ISSUE #288: Use template values for Ingress annotations
Ingress may vary across environments.  I think it would be good to remove 
```
    ingress.kubernetes.io/rewrite-target: /
```
from `templates/ingress.yaml` and relying upon values in `values.yaml`.  This will allow deployment with different ingress implementations to simply provide a values.yaml rather than updating the templates themselves.

prydonius:
This removes the hard dependency on the nginx ingress controller and
enables the use of other ingress controllers

closes #288

### DIFF
--- a/deployment/monocular/Chart.yaml
+++ b/deployment/monocular/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: monocular
 description: Monocular is a search and discovery front end for Helm Charts Repositories.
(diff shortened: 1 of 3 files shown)


prydonius:
cc @sameersbn

rjminsha:
Perfect.  Thanks!

PR #312, opened 2017-07-17
ui: limit shown chart versions to latest 5

### LINKED ISSUE #208: Limit number of versions shown in the UI
We show the whole list of versions when we shown just show the x latest ones.

selection_047

Eventually this should be moved to the API enabling pagination but for now, modifying the UI to only show the latest 5 versions and a link to "view more' would be enough.

You can see an example of the desired behavior here https://rubygems.org/gems/devise

prydonius:
adds a link to show all versions

fixes #208

### DIFF
--- a/src/ui/src/app/chart-details/chart-details-versions/chart-details-versions.component.html
+++ b/src/ui/src/app/chart-details/chart-details-versions/chart-details-versions.component.html
@@ -1,11 +1,14 @@
 <div class="versions">
   <h1>Chart Versions</h1>
-  <div class="version" *ngFor="let version of versions">
+  <div class="version" *ngFor="let version of shownVersions(versions)">
     <span class='number' [class.selected]="isSelected(version)">
       <a [href]="goToVersionUrl(version)" [routerLink]="goToVersionUrl(version)">
         {{ version.attributes.version }}</a>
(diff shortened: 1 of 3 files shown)


src/ui/src/app/chart-details/chart-details-versions/chart-details-versions.component.html
```diff
@@ -1,12 +1,15 @@
...
   <h1>Chart Versions</h1>
-  <div class="version" *ngFor="let version of versions">
+  <div class="version" *ngFor="let version of shownVersions(versions)">
     <span class='number' [class.selected]="isSelected(version)">
       <a [href]="goToVersionUrl(version)" [routerLink]="goToVersionUrl(version)">
         {{ version.attributes.version }}</a>
     </span> -
     <span class='creation-date'>{{ version.attributes.created | date: 'MMM d, y' }}</span>
   </div>
+  <div class="more-link" *ngIf="!showAllVersions">
```
Angelmmiguel: I believe that we should hide the link if there are 5 or less versions for a Chart.

prydonius: ah good catch!

PR #329, opened 2017-07-25
Add support for Traefik as ingress controller

Miouge1:
This adds support for Traefik as an Ingress controller

To use it simply comment the nginx annotations and uncomment the traefik annotations.

Tested with Traefik 1.3.3


prydonius:
Thanks @Miouge1! Do you want this to be released as a new chart version? I'm not sure it makes sense since this is just updating documentation, but if you feel it should then bump the version in Chart.yaml.

Miouge1:
Agreed, I don't think a new version number is necessary either

prydonius:
Ok, in that case I'll merge as-is. Thanks again!

PR #330, opened 2017-07-28
update bitnami/nginx image

prydonius:
- the latest image runs as non-root, and has out of the box support on
  OpenShift


sebgoa:
lgtm

I presume 1.12 is the version that you want, and not the most recent 1.12-master

prydonius:
1.12-master is a dev build, 1.12 is the latest rolling release tag. thanks!

PR #331, opened 2017-07-31
update nginx internal port to 8080

prydonius:
The updated nginx container runs as non-root so needs to run on a non-privileged port


sameersbn:
lgtm!

Issue #257, opened 2017-05-20
Need ability to add non default repos


sebgoa:
Currently we cannot add repos through the UI.

I would like to be able to add a non standard repo.

This is related to authentication and user mgmt

PR #333, opened 2017-08-07
Move repo store from in memory to redis

prydonius:
- Adds Zoom Redis library
- Adds collection for Repos that is backed by Redis
- Removes config.Repo struct and moves to using generated models.Repo


docker-compose.yml
```diff
@@ -22,3 +22,14 @@ services:
       - $HOME/.kube/:/root/.kube
     environment:
       - ENVIRONMENT=development
+  redis:
+    image: bitnami/redis
```
migmartri: Can you use a tagged image?

docker-compose.yml
```diff
@@ -22,3 +22,14 @@ services:
...
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
+    volumes:
+      - 'redis_data:/bitnami/redis'
+    ports:
+      - 6379:6379
```
migmartri: is this port needed to be exposed in the host?

src/api/config/repos/repos.go
```diff
@@ -69,3 +63,7 @@ func loadReposFromFile(filePath string) (Repos, error) {
 	}
 	return yamlStruct.Repos, nil
 }
+
+func strToPtr(s string) *string {
```
migmartri: are you sure this function does not exist somewhere else already?

src/api/data/cache/repos.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/albrow/zoom"
+	"github.com/kubernetes-helm/monocular/src/api/config"
+	"github.com/kubernetes-helm/monocular/src/api/data"
+	"github.com/kubernetes-helm/monocular/src/api/swagger/models"
+)
+
+// Repos is a Zoom Collection for the Repo model
+var Repos *zoom.Collection
+
+// NewCachedRepos returns a data.Repos object to manage repositories
```
migmartri: > returns a data.Repos object to manage repositories

Does it?

src/api/Makefile
```diff
@@ -8,7 +8,7 @@ REPO_PATH := github.com/kubernetes-helm/${SHORT_NAME}
 DEV_ENV_IMAGE := quay.io/deis/go-dev:v0.22.0
 SWAGGER_IMAGE := quay.io/goswagger/swagger:0.6.0
 DEV_ENV_WORK_DIR := /go/src/${REPO_PATH}/src/api
-DEV_ENV_PREFIX := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR}
+DEV_ENV_PREFIX := docker run --rm -e GO15VENDOREXPERIMENT=1 -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR} --net=host
```
migmartri: This is interesting, why is this needed?

docker-compose.yml
prydonius: I'm exposing it because I was testing through my editor so needed to connect from my host, I think it's useful for that especially given testing locally in Go is very common.

(on a side note, I've been considering whether we should remove Docker Compose entirely and move to Minikube with host mounts and something like telepresence.io for the Go service).

src/api/config/repos/repos.go
prydonius: It is in the handlers package, which imports repos so it can't be imported here. I'll move this to a new util package that only holds basic functions like this.

src/api/data/cache/repos.go
prydonius: er.. good catch ;). It did, but I moved to a static variable instead.

src/api/Makefile
prydonius: Rather than change the Makefile to start the test using Docker Compose, I opted to keep this as is and instead make sure a Redis server is running locally. In Travis, I enabled the Redis server and locally I am exporting 6379 from the container. Perhaps this is a bit lazy, and what I should do is change the Makefile to use Compose. What do you think?

prydonius:
> What's the reason for moving repositories to redis? Are we persisting its content across deployments? it seems that configure monocular loads the repos in config on every boot? cache.NewCachedRepos(conf.Repos)

This will lay the groundwork for adding repos at runtime via the UI (see #257), each Monocular instance needs a single source of truth for what repos exist. It uses the config to bootstrap the repos, yes - and actually currently it will add them back if they are deleted at runtime. We might want to change that, but I think it is fine for now. I consider the repos in the config file the "default repos" that should always be configured when starting up - which is useful in a CI/CD environment where you don't want to have to manually configure repos in the UI every time.

> Are we planning on storing the charts as well?

I think this is the next step, yes. Though we may need to consider the points brought up in https://github.com/kubernetes-helm/monocular/issues/251#issuecomment-308812030

src/api/data/util/util.go
```diff
@@ -0,0 +1,12 @@
+// Package util contains small utilities used in other packages
+package util
+
+// Int64ToPtr converts an int64 to an *int64
+func Int64ToPtr(n int64) *int64 {
+	return &n
+}
+
+// StrToPtr converts a string to a *string
+func StrToPtr(s string) *string {
```
mmikulicic: Nitpick `StrToPtrOfWhat?`: obviously of str! hence `StrToStrPtr`, or perhaps just`StrPtr`?

or turning this around: why don't we just call the package `pointerto`, and have `pointerto.String`, `pointerto.Int64`?

src/api/handlers/charts/charts_test.go
```diff
@@ -145,6 +164,27 @@ func TestSearchCharts200(t *testing.T) {
...
+			return ret, errors.New("error searching charts")
+		},
+	})
+	resp := SearchCharts(params, chImplementation)
+	assert.NotNil(t, resp, "SearchCharts response")
+	resp.WriteResponse(w, runtime.JSONProducer())
+	assert.Equal(t, w.Code, http.StatusBadRequest, "expect a 400 response code")
+	var httpBody models.Error
+	assert.NoErr(t, testutil.ErrorModelFromJSON(w.Body, &httpBody))
+	assert.Equal(t, *httpBody.Code, int64(400), "response code in HTTP body data")
```
mmikulicic: is this `int64(..)` really needed for readability? Constant are automatically interpreted to be of the required type.

src/api/config/cache.go
```diff
@@ -0,0 +1,35 @@
+package config
+
+import (
+	log "github.com/Sirupsen/logrus"
+	"github.com/albrow/zoom"
+)
+
+// Pool is a pool of Zoom connections used by other packages
+var Pool *zoom.Pool
```
prydonius: Thanks @mmikulicic, this is definitely the part I was most concerned about with this PR so thanks for the feedback! I agree that `InitPool` is a better name, and I will move to using an accessor with a mutex.

src/api/data/repos.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/kubernetes-helm/monocular/src/api/swagger/models"
+)
+
+// Repo is a Zoom Model for storing repositories
+type Repo models.Repo
+
+// ModelId returns the unique name of the Repo
+func (r *Repo) ModelId() string {
```
prydonius: Yeah, linter complained about it too, but unfortunately this is how they are named in Zoom https://github.com/albrow/zoom#what-is-a-model

src/api/data/util/util.go
prydonius: Whilst creating a package is nice, I can see `util` having more useful shared functions in the future, so I'll rename this to `StrPtr`.

src/api/handlers/charts/charts_test.go
prydonius: This appeared to be failing because it was interpreting `400` as an `int` and not equating it to `*httpBody.Code`, but I'll check again.

src/api/data/util/util.go
mmikulicic: creating packages is cheap.

`util` is a particularly bad name for a package, for a few reasons:

1. by definition, it means that the bunch of thing you put in there has not much in common besides being generally useful. The family of `FooPtr` functions here have a lot in common: they take a Foo and return a *Foo, for the purpose of taking the address of a literal which is not allowed by language spec.

2. without `goimports` programming in Go would be such a pain. `goimports` work by assuming that there is signal in the combination of packagename+function signature. It doesn't work well when a lot of different packages have both the same name and same functions (like the often happens with the various log libraries) and obviously it doesn't work well you need code from multiple packages that are named in the same way (something that usually doesn't happen with the aforementioned log libraries). This makes `util` a choice of last resort, because abusing it will eventually lead to having many `util` packages spread in your code base and in the libraries you depend on and thus a kitten will die every time you'll have to manually import or manually set an alias of an import and/or rename a bunch of package aliases in a code snippet that you're moving around the codebase just because of a conflict with the `util` name. Ah, and some people call it `utils` ... `utils.Foo + util.Bar`

src/api/handlers/charts/charts_test.go
mmikulicic: ah sure, I missed the fact that `assert.Equal` accepts `interface{}`. Yeah, the revolt of the helpers...

src/api/data/util/util.go
prydonius: Fair point, you've convinced me - I'll go for creating a `pointerto` package, I like that it's more readable to. Thanks!

prydonius:
@mmikulicic made all the requested changes - PTAL!

src/api/config/cache.go
```diff
@@ -0,0 +1,56 @@
...
+// GetRedisPool returns a pool of Zoom connections
+func GetRedisPool() *zoom.Pool {
+	once.Do(func() {
+		pool = newRedisPool()
+	})
+	return pool
+}
+
+// CloseRedisPool closes a pool of Zoom connections
+func CloseRedisPool() {
```
prydonius: I was considering adding a mutex here, but I agree that the easier option is to just not close the pool.

prydonius:
There's a bug with the API responses because `getRepoObject` in `handlers` is using the repos from configuration and not from Redis. Since the `handlers` package cannot import the `data/cache` package (import cycle), I need to do some rearranging here.

mmikulicic:
> There's a bug with the API responses because getRepoObject in handlers is using the repos from configuration and not from Redis. Since the handlers package cannot import the data/cache package (import cycle), I need to do some rearranging here.

yeah, lazy initialization is dangerous

src/api/config/repos/repos.go
```diff
@@ -5,33 +5,28 @@ import (
...
-	Name   string
-	URL    string
-	Source string
-}
-
 var official = Repos{
-	Repo{
-		Name:   "stable",
-		URL:    "https://kubernetes-charts.storage.googleapis.com",
+	models.Repo{
```
mmikulicic: I thought we removed these unnecessary type names here

src/api/data/cache/cache.go
```diff
@@ -97,10 +93,16 @@ func (c *cachedCharts) All() ([]*models.ChartPackage, error) {
 	c.rwm.RLock()
 	defer c.rwm.RUnlock()
 	var allCharts []*models.ChartPackage
+	reposCollection, err := data.GetRepos()
+	if err != nil {
+		return nil, err
+	}
+	repos := []*data.Repo{}
```
mmikulicic: I personally prefer to use:

```
var repos []*data.Repo
```

in this case, since it better conveys the idea that it will get filled by `FindAll`

src/api/data/cache/cache.go
```diff
@@ -134,8 +136,14 @@ func (c *cachedCharts) Refresh() error {
 		"path": charthelper.DataDirBase(),
 	}).Info("Using cache directory")
 
-	for _, repo := range c.knownRepos {
-		u, _ := url.Parse(repo.URL)
+	reposCollection, err := data.GetRepos()
+	if err != nil {
+		return err
+	}
+	repos := []*data.Repo{}
```
mmikulicic: same here

PR #337, opened 2017-08-16
show monocular version in ui

### LINKED ISSUE #316: Show version number in the home page
Now that we have a release process,  it would be useful if we had the version number in the home page.

For example, in the footer we have:

"Monocular is an Open Source project"

We could say something

"This site  is running Monocular v0.3.0
Monocular is an Open Source project"

prydonius:
closes #316

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ before_install:
 install:
   - make -C src/api bootstrap
   - make -C src/ui install
+  - |
+    if [[ -n "$TRAVIS_TAG" ]]; then
+      make VERSION=$TRAVIS_TAG set-version
+    fi
 script:
   - make -C src/api test test-cover docker-build
   - make -C src/ui docker-build # Tests disabled for now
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -19,6 +19,10 @@ before_install:
 install:
   - make -C src/api bootstrap
   - make -C src/ui install
+  - |
+    if [[ -n "$TRAVIS_TAG" ]]; then
```
prydonius: cc @sameersbn this will work as intended, right?

sameersbn: I think you meant to write

```
make -C src/ui VERSION=$TRAVIS_TAG set-version
```

sameersbn: As an aside, if you want to git rid of the inline bash from travis, then you could replace the `VERSION ?= dev` assignment in the Makefile with.

```Makefile
ifeq "$(VERSION)" ""
	override VERSION = dev
endif
```

prydonius: oops good catch, thanks

prydonius:
@sameersbn PTAL!

PR #342, opened 2017-08-17
Update Configuring chart repositories section

Locus99:
I believe the first line after the cat command should have a colon after the word 'api'.  Otherwise, it is not proper yaml format.


prydonius:
@Locus99 you're right, how come you closed this?

Locus99:
Did not mean to close this, @prydonius.  Lack of experience on my part.

prydonius:
No problem @Locus99, thanks again :)

PR #347, opened 2017-09-04
Fix service name in logs

osoriano:
Certain server operations (e.g. interrupting the server) cause it to
log "todo list" as the server name. Update these log messages.


prydonius:
Thanks a lot @osoriano!

PR #350, opened 2017-09-05
Add MySQL storage driver

### LINKED ISSUE #348: Add MySQL storage driver
In my environment, I'm using MySQL. I'd like to use a MySQL storage driver instead of the Redis driver that's available today.

osoriano:
* Add a Storage interface that includes operations on Repos
* Include Redis Storage driver based on existing code
* Implement MySQL Storage driver
* Update unit tests to work with both drivers

Fixes #348


prydonius:
I've actually been thinking of moving to either an RDBMS or NoSQL database as we want to move to storing more in database than in memory (e.g. charts currently cached in memory should move to db, when we add authentication it will need to be backed by a db, etc.) Do you have any thoughts on this? Is there a particular reason, other than you have it in your environment, you need to use MySQL?

I'm curious why you opted for using the SQL package directly rather than using an ORM (I've been looking at gorm and others). Whilst the current need for the database is small (only repo CRUD operations), this will extend to other models so I feel an ORM would simplify and remove boilerplate SQL queries. We won't be dealing with any complex queries and performance is not a huge deal, so I think an ORM would be fine to use.

By the way, I really like the Storage interface added in this PR, it will pave the way for adding a mocked implementation that we can use in tests. I don't currently like that we have to spin up and tear down a real database in tests.

.travis.yml
```diff
@@ -11,11 +11,16 @@ cache:
...
+  - mysql
 before_install:
+  # UI dependencies
   - export CHROME_BIN=chromium-browser
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
+  # MySQL Database dependencies
+  - mysql -e 'CREATE DATABASE monocular;'
+  - mysql -e "GRANT ALL PRIVILEGES ON monocular.* To 'monocular'@'localhost' IDENTIFIED BY 'monocular';"
+  - mysql monocular -e "source src/api/storage/mysql/migrations/20170901000000_repos_table.up.sql"
```
prydonius: Just FYI, I've been looking at libraries like https://github.com/pressly/goose and https://github.com/go-gormigrate/gormigrate to manage migrations

src/api/data/helpers/helpers_test.go
```diff
@@ -26,6 +29,23 @@ const (
 	chartHome        = "https://k8s.io/helm"
 )
 
+func TestMain(m *testing.M) {
```
prydonius: why is this test repeated in each package?

src/api/data/repos.go
```diff
@@ -1,47 +1,12 @@
...
 import (
-	"sync"
-
-	"github.com/albrow/zoom"
-	"github.com/kubernetes-helm/monocular/src/api/config"
 	"github.com/kubernetes-helm/monocular/src/api/swagger/models"
 )
 
-// Repo is a Zoom Model for storing repositories
+// Extend the repos model to satisfy the zoom model interface
```
prydonius: AFAIK `go lint` won't like this, can you run your changes through the linter?

src/api/mocks/mocks.go
```diff
@@ -33,6 +33,7 @@ var getTestDataWd = func() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	log.Printf("GOT WD: %s", cwd)
```
prydonius: are these debug statements?

osoriano:
The feedback makes sense. I'll see if I can carve out some time to update the change this week, thanks.

prydonius:
Closing this out for now. Ended up deciding on moving to MongoDB for the database in order to avoid issues with migrations.

Issue #383, opened 2017-10-25
Make API server stateless: move chart data to MongoDB


prydonius:
Currently chart data is stored in memory and fetched via a goroutine at some interval. This is problematic because:

- data is not shared between replicas of the Monocular API server, this means there is the potential for two instances to be out-of-sync
- some data is stored on the filesystem (icons, READMEs) and this can be lost when the pods are restarted
- it's not possible to manually refresh repos if the API server is scaled out (https://github.com/kubernetes-helm/monocular/issues/344#issuecomment-325323237)

The Monocular project is now using a MongoDB data to store repository and user data. This should be extended to charts so we can make the API server truly stateless and therefore more scalable. I propose:

- extracting the repo refresh job from the API server to a standalone tool that writes to a MongoDB database
- run this as a scheduled job (e.g. Kubernetes CronJob)
- a mechanism for kicking off a manual refresh (e.g. service that creates a Kubernetes Job)

prydonius:
One idea for this is to have a controller that tracks a CRD (AppRepositories) that will:
- create a Kubernetes CronJob to perform the sync when an AppRepository is created
- delete a Kubernetes CronJob when an AppRepository is deleted

It's not clear to me how to trigger a manual refresh using the CRD/Controller method. One idea could be to update an arbitrary field in the CRD (e.g. `lastRequestedManualSyncAt: <timestamp>`) which would trigger the controller to run a one-time Job to sync.

A better option might be to create an AppRepositorySyncRequest CRD as well, which we may need to garbage collect after some time.

migmartri:
> create a Kubernetes CronJob to perform the sync when an AppRepository is created
> delete a Kubernetes CronJob when an AppRepository is deleted

WRT using cronjobs, we need to double check its availability in their API servers, for example in <1.8 it is an alpha feature that many clusters might not have enable.

So as an alternative we could make the controller be the one that triggers jobs every x minutes based on a configurable value in the repo CRD.

prydonius:
@migmartri yeah, thanks for pointing it out. I talked about this with @arapulido and our plan is to support N and N-1. Kubernetes 1.9 will be out soon and there's still a few weeks until this change will be put into place, so I think we are okay to use CronJobs.

PR #385, opened 2017-10-26
add chart-repo-sync command

prydonius:
one step towards #383 

This is a command-line tool that takes a repository name and URL and imports chart metadata, icon and README into the database.

Changes from previous sync logic:
- Dumps the repo index into the DB before processing READMEs and icons - this will allow the api server to access the chart metadata before an icon or README is processed for a chart/chart version. This is intentional, the UIs should handle incomplete data (e.g. retrying API call until the README is present).
- We used to fetch an icon for each chart-version pair, chart-repo-sync no longer does this and only processes the latest icon for a chart.
- We process icons and chart tarballs all directly in-memory before saving to the database, nothing is written to the disk. Since the icons and tarballs we're dealing with are small, this shouldn't be an issue.
- We are now doing everything sequentially. This is mostly for simplicity, and should be improved in the future. Specifically, icon and README fetching should be parallelized.


src/api/cmd/chart-repo-sync/sync.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// Syncing is performed in the following sequential steps:
+// 1. Update database to match chart metadata from index
+// 3. Process icons for charts
+// 4. Process READMEs for each chart version
+//
+// These steps are processed in this way to ensure relevant chart data is
+// imported into the database as fast as possible. E.g. we want all icons for
+// charts before fetching readmes for each chart and version pair.
```
prydonius: further explanation: this will enable the UI (e.g. Dashboard) to show all charts as soon as possible, but not have icons and readmes for everything right away. The idea is to have Dashboard startup as quickly as possible.

src/api/cmd/chart-repo-sync/sync.go
```diff
@@ -0,0 +1,335 @@
...
+		return fmt.Errorf("%d %s", res.StatusCode, c.Icon)
+	}
+
+	orig, err := imaging.Decode(res.Body)
+	if err != nil {
+		return err
+	}
+
+	// TODO: make this configurable?
+	icon := imaging.Fit(orig, 160, 160, imaging.Lanczos)
```
prydonius: hardcoding this for now

src/api/cmd/chart-repo-sync/sync.go
```diff
@@ -0,0 +1,335 @@
...
+
+	var b bytes.Buffer
+	imaging.Encode(&b, icon, imaging.PNG)
+
+	db, closer := dbSession.DB()
+	defer closer()
+	return db.C(chartCollection).UpdateId(c.ID, bson.M{"$set": bson.M{"raw_icon": b.Bytes()}})
+}
+
+func fetchAndImportReadmes(c chart) {
```
prydonius: this needs to be improved, we're doing it sequentially for now. However, given the order of how we import things has now changed, I think this is okay.

migmartri:
> Dumps the repo index into the DB before processing READMEs and icons - this will allow the api server to access the chart metadata before an icon or README is processed for a chart/chart version. 

I do not agree with this statement. In my opinion the UI should never receive incomplete charts, it should be able to start populating the ones that have been completed though (which I believe that it is not the case today because of the way we populate the whole repo struct before returning it)

> This is intentional, the UIs should handle incomplete data (e.g. retrying API call until the README is present).

This seems an overkill in my opinion. And if in the end we do this, I think we should return some kind of `completed` flag so the UI does not need to have logic to know what completed means. In any case, I'd make the backend be in charge of this as I stated in my previous comment.

migmartri:
> We process icons and chart tarballs all directly in-memory before saving to the database, nothing is written to the disk. Since the icons and tarballs we're dealing with are small, this shouldn't be an issue.

Currently Readme and Icons are provided using an static files webserver which adds the flexibility of the possibility of adding caching and of course not hitting the API.

What's our plan for exposing these static assets from the database all the way to the client? I was wondering if we could still write files to disk so the API is the same but if we miss that cache we populate that file from the DB.

src/api/cmd/chart-repo-sync/sync.go
migmartri: As I said in my previous comment, in my opinion, I would only expose the charts that have been fully imported, but the UI will start showing charts as soon as 1 chart has been fully processed, which is not the case today.

src/api/cmd/chart-repo-sync/sync.go
migmartri: Doesn't the old importer executes with a pool of x parallel workers? Could we implement that here? It seems that not much has changed in the process that will prevent us to have that importation boost from day one?

src/api/cmd/chart-repo-sync/sync_test.go
```diff
@@ -0,0 +1,447 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/arschles/assert"
+	"github.com/disintegration/imaging"
+	"github.com/kubeapps/common/datastore/mockstore"
+	"github.com/stretchr/testify/mock"
+	"gopkg.in/mgo.v2/bson"
+)
+
+var validRepoIndexYAML = `
+entries:
```
migmartri: Could you store this as another file inside a `testdata` directory?

src/api/cmd/chart-repo-sync/sync_test.go
```diff
@@ -0,0 +1,447 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := sync("test", tt.repoURL)
+			assert.ExistsErr(t, err, tt.name)
+		})
+	}
+}
+
+func Test_fetchRepoIndex(t *testing.T) {
```
migmartri: I am curious, why are you using the underscore after the Test required word? Is this a new GO idiom?

src/api/cmd/chart-repo-sync/sync_test.go
```diff
@@ -0,0 +1,447 @@
...
+func Test_fetchAndImportIcon(t *testing.T) {
+	t.Run("no icon", func(t *testing.T) {
+		c := chart{ID: "test/acs-engine-autoscaler"}
+		assert.NoErr(t, fetchAndImportIcon(c))
+	})
+
+	index, _ := parseRepoIndex([]byte(validRepoIndexYAML))
+	charts := chartsFromIndex(index, repo{Name: "test", URL: "http://testrepo.com"})
+
+	t.Run("failed download", func(t *testing.T) {
```
migmartri: Very cool use the subtest cases (t.Run) I was not aware of it :)

src/api/cmd/chart-repo-sync/sync_test.go
prydonius: +1 will move it

src/api/cmd/chart-repo-sync/sync_test.go
prydonius: I believe so for internal/unexposed function names. VSCode automatically creates these names.

src/api/cmd/chart-repo-sync/sync_test.go
prydonius: new feature in 1.7, I really like it!

src/api/cmd/chart-repo-sync/sync.go
prydonius: Well, yes, but I wanted to get a first cut of this and decided to avoid implementing the worker pool straight away. How about doing it in a separate PR after this? This won't be put into use right away (the API has not been updated to serve from the database yet), so in terms of making the review process easier we can do it in a follow-up PR. Just FYI, I found that the old importer creates n goroutines (when n = number of charts) and then a buffered channel is setup as a semaphore to only allow x (some number of) goroutines to execute at a time. I think we should rework this so that we only create x goroutines and process it like a worker pool e.g. see https://gobyexample.com/worker-pools

prydonius:
> I do not agree with this statement. In my opinion the UI should never receive incomplete charts, it should be able to start populating the ones that have been completed though (which I believe that it is not the case today because of the way we populate the whole repo struct before returning it)

> This seems an overkill in my opinion. And if in the end we do this, I think we should return some kind of completed flag so the UI does not need to have logic to know what completed means. In any case, I'd make the backend be in charge of this as I stated in my previous comment.

What do you see as a complete chart, though? In my view, all we need for a functional Dashboard is the ability to view a list of charts and install them. Perhaps seeing the README is important for that, but I wouldn't say the icons are mandatory (they will eventually appear, of course). One idea is that we fetch the README for the latest chart first, that way you at least have the main page of a chart working right away. Clicking on older versions of a chart that have not been loaded can show a loading screen.

I think a completed flag could be useful, I'll have to think about how to handle that. Alternatively, we should be able to expose whether a repo sync is currently running or not (e.g. querying the k8s api for incomplete k8s jobs), and the UI can switch on polling whilst that's happening.

> Currently Readme and Icons are provided using an static files webserver which adds the flexibility of the possibility of adding caching and of course not hitting the API.
>
> What's our plan for exposing these static assets from the database all the way to the client? I was wondering if we could still write files to disk so the API is the same but if we miss that cache we populate that file from the DB.

Good point, my plan was to do the simplest thing first and just live-read this from the DB. The client (browser) could handle caching of these files. If we do need to cache this, we could fetch from the DB the first time and download the file and the next time have the static file server serve it. But I think we can certainly start with the simplest approach first.

PR #389, opened 2017-11-07
Adds GH login to user collection

sameersbn:
**Changes**
- oauth: save github login id to user collection
- oauth: lookup the gh login id while registering/logging in a user
  Unlike the primary email the github login id is immutable and is
  therefore a better candidate to perform a user lookup


sameersbn:
Does not seem like a good idea considering we may want to support oauth from various other services. Closing...

PR #396, opened 2017-11-21
oauth: fallback to using the gh username if Name is not defined

sameersbn:


prydonius:
@sameersbn do we still want this?

sameersbn:
Yes. I think we need this.

The issue is that github makes the full name of a user optional. The hub ui associates the user's full name to comments and when not specified in the gh profile, no name will be associated with comments.

The idea of this PR is to fallback to using the gh username when the full name is not specified by a user in their GH profile. Alternatively we could display some placeholder text if we choose not to merge this.

PR #400, opened 2017-11-24
set user agent when requesting from chart repos

### LINKED ISSUE #394: Set user agent when pulling chart tarballs and repo indexes
In order for us to get more reliable chart download data for the stable repos we want to be able to filter by user agents. We should fix Monocular so that we're always using a custom user agent and not polluting the chart download data.

cc @viglesiasce

prydonius:
fixes #394


prydonius:
cc @mattfarina @viglesiasce

PR #403, opened 2017-11-30
add optional annotations to api and ui kubernetes  services

rawlingsj:
this helps when running controllers that watch services and can react to annotations


prydonius:
Thanks @rawlingsj!!

PR #411, opened 2017-12-21
chart-repo-sync: concurrently process charts

prydonius:
Implements a worker pool (hardcoded to 10 workers) to download and process chart icons and tarballs concurrently.

Also changed to process the latest chart tarball of each chart first before going on to process historical chart versions.

### DIFF
--- a/src/api/cmd/chart-repo-sync/sync.go
+++ b/src/api/cmd/chart-repo-sync/sync.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -29,6 +30,12 @@ const (
 	chartCollection       = "charts"
 	chartReadmeCollection = "readmes"
 )
+
+type importChartFilesJob struct {
+	Name         string
+	Repo         repo
+	ChartVersion chartVersion
+}
 
 type httpClient interface {
 	Do(req *http.Request) (*http.Response, error)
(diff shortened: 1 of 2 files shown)


src/api/cmd/chart-repo-sync/sync.go
```diff
@@ -100,15 +108,47 @@ func sync(repoName, repoURL string) error {
 		return err
 	}
 
+	// Process 10 charts at a time
+	numWorkers := 10
```
sameersbn: do you think making this a configurable parameter be a good idea?

prydonius: I was thinking about it, yeah. I think for now this is OK to hardcode, we can make it configurable if there becomes a need for that later

src/api/cmd/chart-repo-sync/sync.go
```diff
@@ -68,21 +75,22 @@ func main() {
...
 		os.Exit(1)
 	}
 }
 
-// Syncing is performed in the following sequential steps:
+// Syncing is performed in the following steps:
 // 1. Update database to match chart metadata from index
-// 3. Process icons for charts
-// 4. Process READMEs for each chart version
+// 3. Concurrently process icons for charts (concurrently)
```
migmartri: I do not fully understand the numbering here.

src/api/cmd/chart-repo-sync/sync.go
```diff
@@ -100,15 +108,47 @@ func sync(repoName, repoURL string) error {
...
+
+	// Enqueue jobs to process chart icons
 	for _, c := range charts {
-		if err := fetchAndImportIcon(c); err != nil {
-			log.WithFields(log.Fields{"name": c.Name}).WithError(err).Error("failed to import icon")
+		iconJobs <- c
+	}
+	// Close the iconJobs channel to signal the worker pools to move on to the
+	// chart files jobs
+	close(iconJobs)
```
migmartri: I am not that familiar with GO anymore but does this mean that we wait to import the chartFiles until all the icons for all the charts have been downloaded?

prydonius: Yes, that's right. Do you think we should get the icon and the readme for the first chart at the same time? I'm not sure it makes too much of a difference tbh.

src/api/cmd/chart-repo-sync/sync.go
prydonius: Apparently I don't know how to count, will fix.

src/api/cmd/chart-repo-sync/sync.go
migmartri: Well, if you show the charts even without readmes and icons I do not think that it makes a difference. If you only expose in the API fully imported charts, making it fully parallel will be good.

Because I think that the implemented the former, this looks good to me.

Thanks for the explanation.

src/api/cmd/chart-repo-sync/sync.go
migmartri: I understand you, I only know how to count in roman numbers :trollface:

migmartri:
wrt to the coverage report. If time permits it would be good if we keep it green :)

src/api/cmd/chart-repo-sync/sync.go
prydonius: @migmartri it sounds like what you're suggesting is an ordering issue and not necessarily the parallelisation. With this approach, it is possible that an icon is processed at the same time as a chart's readmes, it's just that we queue up all the icons before the chart readmes so this only happens when there are < 10 icons left to process.

What would be a better way of queuing the jobs? It's somewhat more complicated by the fact that we are choosing the process the latest readme for each chart first before historical ones. However, I guess a good improvement would be to queue icon and latest chart readme at the same time.

prydonius:
@migmartri r.e. coverage, it's difficult, I've tried to separate a lot of stuff into separate testable functions, it's the main functions that call these that aren't tested. It would be difficult to fully test syncRepo due to the concurrency, for example.

prydonius:
Merging for now, we can look into this again later

PR #413, opened 2017-12-22
kubernetes-helm/monocular#412 Add Values in API

sebadiaz:
Add a map to push on api and helm client

### DIFF
--- a/src/api/swagger/restapi/operations/releases/create_release.go
+++ b/src/api/swagger/restapi/operations/releases/create_release.go
@@ -87,7 +87,7 @@ type CreateReleaseBody struct {
 
 	/* Values of the new release
 	 */
-	Values map[string]string `json:"values,omitempty"`
+	Values string `json:"values,omitempty"`
 }
 
 // Validate validates this create release body
--- a/src/api/swagger/restapi/operations/releases/create_release.go
+++ b/src/api/swagger/restapi/operations/releases/create_release.go
@@ -85,7 +85,7 @@ type CreateReleaseBody struct {
 	 */
 	ReleaseName string `json:"releaseName,omitempty"`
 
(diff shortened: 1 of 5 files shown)


thelinuxfoundation:

* One or more commits are not associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub user profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,
CLA GitHub bot_

src/api/swagger/restapi/operations/releases/create_release.go
```diff
@@ -84,6 +84,10 @@ type CreateReleaseBody struct {
 	/* Name of the new release
 	 */
 	ReleaseName string `json:"releaseName,omitempty"`
+
+	/* Values of the new release
+	 */
+	Values map[string]string `json:"values,omitempty"`
```
prydonius: This will only deal with top-level values. What if we just make this a raw string and just assume it is correctly formed. The Helm client takes a string anyway, so we wouldn't need to convert the map to a string, and it should return an error if this is badly formed.

sebadiaz: I would do that . But how to prevent or escape values including = and , ???

sebadiaz:
Check cla

sebadiaz:
I have moved the map to a string

prydonius:
@sebadiaz looks good, thanks. Were you able to get the CLA signed? Looks like it's not getting picked up. Checkout https://github.com/kubernetes/kubernetes/wiki/CLA-FAQ if you need some help.

sebadiaz:
moved to https://github.com/kubernetes-helm/monocular/pull/415

PR #421, opened 2018-01-30
Add support for extracting values.yaml from charts for chart-repo-sync

sozercan:


thelinuxfoundation:

* User @sozercan isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

sozercan:
I signed it

prydonius:
@sozercan I think it would be better to generalise the readme/values fetching. In the current approach, we're downloading the tarball twice for charts that don't already have the readme or values indexed, so one option is to extract the tarball downloading to a separate function and only download it once.

A better option might be to rename `fetchAndImportReadme` to `fetchAndImportFiles` and extract both the readme and values at the same time. I also think it makes sense to change the MongoDB collection to `files` and store the values as just another field in the same MongoDB document.

PR #425, opened 2018-03-06
Update helm/monocular to kubernetes-helm/monocular

jlegrone:
This gets the TLDR instructions working.


prydonius:
oops, thanks!

PR #442, opened 2018-04-30
fix: work around of regression in helm 2.9.0

### LINKED ISSUE #441: monocular chart broken on 2.9.0 of helm
we just saw this on Jenkins X. We've been using 0.5.0 of monocular and its working fine. But as soon as folks move to 2.9.0 of helm it breaks badly - the API pod keeps failing.

Turns out its an effect of helm 2.9.0 where this line: 
https://gist.github.com/jstrachan/9e31d54766d3eb20bbf63fde5990b902#file-api-config-yaml-L12

no longer generates a newline like this:
https://gist.github.com/rawlingsj/4be1f93cb7943971a020146b13d31de9#file-gistfile1-txt-L16-L17

and so generates this:
https://gist.github.com/jstrachan/a8228416b0ee07056984baa39ca15111#file-configmap-yaml-L16
which fails to parse (as there's no newline between `tillerNamespace: kube-system` and `mongodb:`

A simple fix could be to move the this line https://gist.github.com/jstrachan/9e31d54766d3eb20bbf63fde5990b902#file-api-config-yaml-L12 after the mongo settings?

jstrachan:
this should hopefully work around a regression we've just seen in helm 2.9.0

fixes #441

### DIFF
--- a/deployment/monocular/templates/api-config.yaml
+++ b/deployment/monocular/templates/api-config.yaml
@@ -9,7 +9,7 @@ metadata:
     heritage: "{{ .Release.Service }}"
 data:
   monocular.yaml: |-
-{{ toYaml .Values.api.config | indent 4 -}}
     mongodb:
       url: {{ template "mongodb.fullname" . }}:27017
       database: monocular
+{{ toYaml .Values.api.config | indent 4 -}}


prydonius:
@jstrachan sorry for the delay, I somehow missed this! Thanks a lot!

PR #447, opened 2018-05-09
refresh single chart draft

samifruit514:
Hello there,

I made some changes to support the refresh of a single chart. With this feature, we are able to upload a chart into a chart repo and trigger the refresh in monocular automatically right after, in a programmatic way. To resume, this provides a way to see the chart in monocular immediately after it was uploaded in the repo.

an example call to the api would be: 
curl -X POST http://localhost:8080/v1/charts/incubator/zookeeper/refresh

This is a first draft. If the feature is ok for you, I can proceed with polishing the PR's code.

Thanks

### DIFF
--- a/src/api/data/cache/cache.go
+++ b/src/api/data/cache/cache.go
@@ -2,10 +2,6 @@ package cache
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"net/url"
-	"path"
 	"strings"
 	"sync"
 
@@ -13,12 +9,12 @@ import (
 
 	"github.com/kubernetes-helm/monocular/src/api/data"
 	"github.com/kubernetes-helm/monocular/src/api/data/cache/charthelper"
+	"github.com/kubernetes-helm/monocular/src/api/data/cache/repohelper"
 	"github.com/kubernetes-helm/monocular/src/api/data/helpers"
 	"github.com/kubernetes-helm/monocular/src/api/datastore"
 	"github.com/kubernetes-helm/monocular/src/api/models"
 	swaggermodels "github.com/kubernetes-helm/monocular/src/api/swagger/models"
 	"github.com/kubernetes-helm/monocular/src/api/swagger/restapi/operations/charts"
-	"github.com/kubernetes-helm/monocular/src/api/version"
 )
 
 type cachedCharts struct {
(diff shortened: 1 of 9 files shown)


thelinuxfoundation:

* One or more commits are not associated with a GitHub user. The email used in the Git commit must match a verified email address in a GitHub user profile. For more information, please see https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/

_Regards,
CLA GitHub bot_

prydonius:
this is great, thanks @samifruit514! I'm wondering though if it makes sense to only do this for a single chart? Wouldn't it make more sense to refresh the repo as a whole? Otherwise we're fetching the index but ignoring every other chart.

I can see this might be useful in larger repositories though, as downloading the missing tarballs might take some time otherwise.

src/api/data/cache/cache.go
```diff
@@ -133,6 +133,82 @@ func (c *cachedCharts) Search(params charts.SearchChartsParams) ([]*swaggermodel
 	return ret, nil
 }
 
+// Refresh is the interface implementation for data.Charts
+// It refreshes cached data for a specific repo and chart
+func (c *cachedCharts) RefreshChart(repoName string, chartName string) error {
```
prydonius: we can probably move some of these things into helper functions

samifruit514:
Thanks for your quick feedback @prydonius ! Indeed the whole yaml file is fetched for a single chart update. Short term, the only part we would win is when it verifies if the chart folder exists which is trivial I think...

Long term, if some day the helm repo spec allows to fetch specific charts from the big index, this would make more sense and this refresh route would be already setup.

In my last commit, I extracted common code from the Refresh and RefreshChart to put it in a new helper package I added: repohelpers

Let me know what you think!
Thanks a lot!

src/api/data/cache/repohelper/repo_helper.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/kubernetes-helm/monocular/src/api/data/helpers"
+	"github.com/kubernetes-helm/monocular/src/api/models"
+	swaggermodels "github.com/kubernetes-helm/monocular/src/api/swagger/models"
+	"github.com/kubernetes-helm/monocular/src/api/version"
+)
+
+// GetRepoIndexFile Get the charts from the index file
+var GetChartsFromRepoIndexFile = func(repo *models.Repo) ([]*swaggermodels.ChartPackage, error) {
+
```
prydonius: remove trailing line

prydonius:
Thanks a lot @samifruit514!

PR #476, opened 2018-07-17
Moved the import path to be in the form of github.com/helm/monocular

mattfarina:
This is part of the move to the new org as Monocular as part of Helm
is a top level CNCF project


prydonius:
I guess travis will only pass once this is actually moved, we can keep it open and merge shortly after the move. WDYT?

PR #509, opened 2018-09-13
adds chartsvc deployment to the chart

### LINKED ISSUE #500: Add chartsvc for chart metadata serving API
Add chartsvc to be deployed with the Monocular chart and internally exposed via a Service.

prydonius:
chartsvc is intended to replace the existing Monocular API server for serving the chart metadata being synced using the chart-repo tool over a REST API. It implements the same routes as the Monocular API server for charts and chart versions: https://github.com/helm/monocular/blob/master/src/api/main.go#L81-L91

fixes #500

### DIFF
--- a/deployment/monocular/templates/chartsvc-deployment.yaml
+++ b/deployment/monocular/templates/chartsvc-deployment.yaml
@@ -36,6 +36,10 @@ spec:
         ports:
         - name: http
           containerPort: {{ .Values.chartsvc.service.port }}
+        livenessProbe:
+{{ toYaml .Values.chartsvc.livenessProbe | indent 10 }}
+        readinessProbe:
+{{ toYaml .Values.chartsvc.readinessProbe | indent 10 }}
         resources:
 {{ toYaml .Values.chartsvc.resources | indent 12 }}
     {{- with .Values.chartsvc.nodeSelector }}
--- a/deployment/monocular/templates/chartsvc-deployment.yaml
+++ b/deployment/monocular/templates/chartsvc-deployment.yaml
@@ -8,7 +8,7 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 spec:
-  replicas: 1
+  replicas: {{ .Values.chartsvc.replicas }}
   selector:
     matchLabels:
       app: {{ template "fullname" . }}-chartsvc
--- a/deployment/monocular/templates/chartsvc-deployment.yaml
+++ b/deployment/monocular/templates/chartsvc-deployment.yaml
@@ -1,0 +1,52 @@
+apiVersion: apps/v1beta2
+kind: Deployment
+metadata:
+  name: {{ template "fullname" . }}-chartsvc
+  labels:
+    app: {{ template "fullname" . }}-chartsvc
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ template "fullname" . }}-chartsvc
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "fullname" . }}-chartsvc
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+      - name: chartsvc
+        image: {{ template "monocular.image" .Values.chartsvc.image }}
+        command:
+        - /chartsvc
+        args:
+        - --mongo-user=root
+        - --mongo-url={{ template "mongodb.fullname" . }}
+        env:
+        - name: MONGO_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "mongodb.fullname" . }}
+              key: mongodb-root-password
+        ports:
+        - name: http
+          containerPort: {{ .Values.chartsvc.service.port }}
+        resources:
+{{ toYaml .Values.chartsvc.resources | indent 12 }}
+    {{- with .Values.chartsvc.nodeSelector }}
+      nodeSelector:
+{{ toYaml . | indent 8 }}
+    {{- end }}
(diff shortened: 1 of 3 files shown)


deployment/monocular/values.yaml
```diff
@@ -14,6 +14,25 @@ sync:
     #- name: my-repo-name
     #  url: my-repository-url
     #  source: my-repository-source
+  resources: {}
```
migmartri: We need to revisit later the resources setup since monocular has proper resources settings, so we should expect to do the same with these new services. 

We can work on that in parallel in Kubeapps https://github.com/kubeapps/kubeapps/issues/478 but it would be good if you add to your TODO list to apply whatever values we came up with. Thanks.

deployment/monocular/values.yaml
```diff
@@ -14,6 +14,25 @@ sync:
...
+  nodeSelector: {}
+  tolerations: []
+  affinity: {}
+
+# Chartsvc is used to serve chart metadata over a REST API.
+chartsvc:
+  image:
+    registry: docker.io
+    repository: kubeapps/chartsvc
+    tag: v1.0.0-beta.0
```
migmartri: I do not remember but did you end up setting a fixed tag in the appRepo controller?

deployment/monocular/templates/chartsvc-deployment.yaml
```diff
@@ -0,0 +1,52 @@
...
+kind: Deployment
+metadata:
+  name: {{ template "fullname" . }}-chartsvc
+  labels:
+    app: {{ template "fullname" . }}-chartsvc
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  replicas: 1
```
migmartri: BTW, I believe that monocular might have been using more than 1 replica. We should revisit having more than 1 replica all the time before doing the switch.

deployment/monocular/values.yaml
prydonius: The current best practices are to leave resources blank rather than impose requests/limits, as usually this is best determined by the cluster operator for their environment.

deployment/monocular/values.yaml
prydonius: for the chart-repo CLI, yes.

deployment/monocular/templates/chartsvc-deployment.yaml
migmartri: I have created also a task to tackle this in Kubeapps so we can double check that there is no any architectural issue on doing so https://github.com/kubeapps/kubeapps/issues/643

prydonius: True, do you think we should have 3 chartsvc deployments? MongoDB is single-node and with chartsvc relying heavily on that, that's more of a source of failure. That said, it still might make sense. WDYT?

deployment/monocular/values.yaml
migmartri: > The current best practices are to leave resources blank rather than impose requests/limits, as usually this is best determined by the cluster operator for their environment.

I do not agree with this, there is nobody better to determine the minimum required resources than the app developer. Setting limits == requests will force the scheduler to run it in QOS and not to be affected by the noise in the cluster. Being in the best interest of the developer itself.

I am ok if we do not want to enforce it, but we should at least suggest it. Although TBH in my opinion, no app should be run in production without requests/limits unless it is designed for using of residual compute.

deployment/monocular/templates/chartsvc-deployment.yaml
migmartri: > True, do you think we should have 3 chartsvc deployments? MongoDB is single-node and with chartsvc relying heavily on that, that's more of a source of failure

My goal is to not to have 1. 2 is fine. I understand your point about MongoDB but I think that's a bigger issue and actually Monocular is not solving it either yet. Also MongoDB can always be replaced for a managed database while our services aren't.

In the long term I agree that the MongoDB bottleneck needs to be resolved to avoid cascading failures.

deployment/monocular/values.yaml
prydonius: Sorry, I thought this was written down in the Helm docs, but I can't find it so may not actually be an agreed best practice. I do remember having a discussion over it, though, and the decision was to not set anything by default (`helm create` keeps it empty, as do most charts).

I see what you're saying, and I do think that it's useful to provide an example, but IMO it's not really possible to make a good judgement without also knowing what the available resources are and what else is running in the cluster. That may be less important for a small service like this, though.

Given that we don't have a good idea of the resource usage for chartsvc today, I think we should leave this as is and revisit it once kubeapps/kubeapps#478 is complete.

deployment/monocular/templates/chartsvc-deployment.yaml
prydonius: I believe it's more common to use 3 replicas (odd numbers), but I am fine with either. As for MongoDB, agreed, and I'm looking into whether the Helm Hub can use CosmosDB.

prydonius:
@migmartri I think I addressed all your comments, PTAL!

deployment/monocular/values.yaml
migmartri: > helm create keeps it empty, as do most charts).

That makes sense since it is the developer the one that will tweak it based on the app that it's packaging?

> I see what you're saying, and I do think that it's useful to provide an example, but IMO it's not really possible to make a good judgement without also knowing what the available resources are and what else is running in the cluster. 

Why is that? Can't we figure out what are the minimum requirements for these component? The community already defined them for the previous components https://github.com/helm/monocular/blob/master/deployment/monocular/values.yaml#L17

Note, I am talking about reasonable "minimum" requirements.

> Given that we don't have a good idea of the resource usage for chartsvc today, I think we should leave this as is and revisit it once kubeapps/kubeapps#478 is complete.

Sounds good, thanks!

migmartri:
> @migmartri I think I addressed all your comments, PTAL!

Thanks, It looks good.

PR #511, opened 2018-09-18
Updates the Monocular UI to use the chartsvc API

### LINKED ISSUE #501: Update Monocular UI to use chartsvc
Replace API calls to the Monocular API server to use the chartsvc API.

prydonius:
In https://github.com/helm/monocular/pull/509 we added the chartsvc
service to the Monocular distribution, intended to be a lighter-weight,
read-only replacement for the Monocular API server. This PR updates the
Monocular UI to switch to the new chartsvc API for all chart-related API
calls.

fixes #501

### DIFF
--- a/deployment/monocular/templates/ingress.yaml
+++ b/deployment/monocular/templates/ingress.yaml
@@ -25,6 +25,10 @@ spec:
           serviceName: {{ template "fullname" $ }}-api
           servicePort: {{ $.Values.api.service.externalPort }}
         path: /api/
+      - backend:
+          serviceName: {{ template "fullname" $ }}-chartsvc
+          servicePort: {{ $.Values.chartsvc.service.port }}
+        path: /api/chartsvc
     host: {{ . | quote }}
   {{- end }}
 {{- if .Values.ingress.tls }}
--- a/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
+++ b/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
@@ -24,21 +24,8 @@ export class ChartDetailsInfoComponent implements OnInit {
     return this.chart.attributes.sources || [];
   }
 
-  get sourceUrl(): string {
-    var chartSource = this.chart.attributes.repo.source;
-    if (!chartSource) return;
-
-    return urljoin(chartSource, this.chart.attributes.name);
-  }
-
   get maintainers(): Maintainer[] {
     return this.chart.attributes.maintainers || [];
-  }
-
-  get sourceName(): string {
-    var parser = document.createElement('a');
-    parser.href = this.chart.attributes.repo.source;
-    return parser.hostname;
   }
 
   loadVersions(chart: Chart): void {
--- a/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
+++ b/src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
@@ -50,14 +50,16 @@ export class ChartDetailsInfoComponent implements OnInit {
   }
 
   maintainerUrl(maintainer: Maintainer): string {
-    if (this.isKubernetesCharts(this.chart.attributes.repo.URL)) {
+    // Use GitHub URL with maintainer name if this is an upstream Helm repo from
+    // github.com/helm/charts (i.e. stable or incubator)
+    if (this.isUpstreamHelmRepo(this.chart.attributes.repo.url)) {
       return `https://github.com/${maintainer.name}`;
     } else {
       return `mailto:${maintainer.email}`;
     }
   }
 
-  private isKubernetesCharts(repoURL: string): boolean {
+  private isUpstreamHelmRepo(repoURL: string): boolean {
     return (
       repoURL === "https://kubernetes-charts.storage.googleapis.com" ||
       repoURL === "https://kubernetes-charts-incubator.storage.googleapis.com"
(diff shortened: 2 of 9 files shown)


deployment/monocular/templates/ingress.yaml
```diff
@@ -25,6 +25,10 @@ spec:
           serviceName: {{ template "fullname" $ }}-api
```
migmartri: Maybe add a note saying that this backend is deprecated?

src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
```diff
@@ -50,10 +50,17 @@ export class ChartDetailsInfoComponent implements OnInit {
...
   maintainerUrl(maintainer: Maintainer): string {
-    if (this.chart.attributes.repo.source.match(/github.com/)) {
+    if (this.isKubernetesCharts(this.chart.attributes.repo.URL)) {
       return `https://github.com/${maintainer.name}`;
     } else {
       return `mailto:${maintainer.email}`;
     }
   }
+
+  private isKubernetesCharts(repoURL: string): boolean {
```
migmartri: The signature is a little bit confusing to me. Maybe, isUpstreamChartRepo?

src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
```diff
@@ -50,10 +50,17 @@ export class ChartDetailsInfoComponent implements OnInit {
   }
 
   maintainerUrl(maintainer: Maintainer): string {
-    if (this.chart.attributes.repo.source.match(/github.com/)) {
+    if (this.isKubernetesCharts(this.chart.attributes.repo.URL)) {
```
migmartri: So let me understand the assumptions made here.

Before, a source was set in the repo info. Now we assume that the only two repositories that we know for sure that have a github repo are the upstream one no?

If that's the case, I'll add a comment explaining the assumption

src/ui/src/app/chart-details/chart-details.component.ts
```diff
@@ -71,12 +71,8 @@ export class ChartDetailsComponent implements OnInit {
   }
 
   getIconUrl(): string {
-    let icons = this.chart.relationships.latestChartVersion.data.icons;
-    if (icons !== undefined && icons.length > 0) {
-      const icon =
-        this.config.backendHostname +
-        icons.find(icon => icon.name === '160x160-fit').path;
-      return icon;
+    if (this.chart.attributes.icon) {
```
migmartri: Nice :tada:

src/ui/src/app/shared/services/charts.service.ts
```diff
@@ -127,6 +127,19 @@ export class ChartsService {
       .catch(this.handleError);
   }
 
+  /**
+   * Get the URL for retrieving the chart's icon
+   * 
+   * @param {string} chart Chart object
```
migmartri: How does this annotation work? {string} is the returned value? What's the name of this annotation?

src/ui/src/app/shared/services/charts.service.ts
```diff
@@ -127,6 +127,19 @@ export class ChartsService {
...
+   */
+  getChartIconURL(chart: Chart): string {
+    if (chart.attributes.icon) {
+      return `${this.hostname}${chart.attributes.icon}`
+    } else {
+      return null;
```
migmartri: Maybe you can return here the placeholder so the caller does not do the check also?

migmartri: Which seems to happen in multiple places.

deployment/monocular/templates/ingress.yaml
prydonius: Perhaps not really necessary since this is going to be removed in the next couple of PRs, and this is on the dev-v1 branch so it won't affect any current users. WDYT?

src/ui/src/app/chart-details/chart-details-info/chart-details-info.component.ts
prydonius: Yes, that's the assumption we're making here. Will add a comment!

src/ui/src/app/shared/services/charts.service.ts
prydonius: It has no semantic meaning, but good catch, that should actually be the type of the parameter which is a Chart.

src/ui/src/app/shared/services/charts.service.ts
prydonius: I was thinking about it, but the placeholder is not something from the chartsvc, but part of the Angular app, so it might be a bit weird. However, it would reduce boilerplate so maybe it's ok to put it here.

deployment/monocular/templates/ingress.yaml
migmartri: You are right, but not always assume that you should only be the one that has the context of what's next. Adding that comment, even if temporary could help others to continue your work and for sure to not to forget.

src/ui/src/app/shared/services/charts.service.ts
migmartri: The chartSVC is a consumer class in the frontend no? Not something that we import/re-use from the chart-svc? In which case I think it's ok to move the placeholder here. 

I'd understand that we wouldn't want to do it if we had different placeholders. Up to you though.

prydonius: Yes it is, but I figured it should keep the context of talking to the chartsvc, but I agree it doesn't make much sense to have the if/else logic repeated everywhere we want to use it. Will move it here then. Thanks!

prydonius:
@migmartri I believe I've addressed all your comments, PTAL!

PR #513, opened 2018-09-18
remove package-lock.json

### LINKED ISSUE #512: UI: Remove package-lock.json in favour of yarn.lock
Yarn complains during install due to the existence of the package-lock.json file which can potentially conflict with yarn.lock. We should remove this file since we use yarn.lock instead.

prydonius:
Since we use yarn for the Angular app, which uses yarn.lock, we should
not have this confliciting package-lock.json file.

fixes #512


prydonius:
Yeah the makefile is using yarn exclusively, thanks!

PR #516, opened 2018-09-18
ci: enable dev-v1 branch testing

prydonius:
We were previously only enabling CI builds on the master branch and
tags. This updates the Travis CI configuration to add the dev-v1 branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ before_install:
 install:
   - make -C src/ui install
   - make -C src/ui VERSION=$TRAVIS_TAG set-version
+  # required due to apparent yarn issue with node-sass: https://github.com/sass/node-sass/issues/1971
+  - npm --prefix src/ui rebuild node-sass
 script:
   - make -C src/ui docker-build # Tests disabled for now
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-# Use older version of Trusty VM due to issues in the updated one
-group: deprecated-2017Q2
-filter_secrets: false
 language: node_js
 node_js:
   - '8.1.4'
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ script:
   - make -C src/ui docker-build # Tests disabled for now
 after_success:
   - |
-    if [[ "$TRAVIS_PULL_REQUEST" = "false" ]] && [[ "$TRAVIS_BRANCH" = "master" || -n "$TRAVIS_TAG" ]]; then
+    if [[ "$TRAVIS_PULL_REQUEST" = false ]] && [[ "$TRAVIS_BRANCH" = "master" || -n "$TRAVIS_TAG" ]]; then
       if [[ -n "$TRAVIS_TAG" ]]; then
         docker tag bitnami/monocular-ui:latest bitnami/monocular-ui:$TRAVIS_TAG
       fi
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -26,7 +26,7 @@ script:
   - make -C src/ui docker-build # Tests disabled for now
 after_success:
   - |
-    if [[ "$TRAVIS_PULL_REQUEST" == false ]]; then
+    if [[ "$TRAVIS_PULL_REQUEST" = "false" ]] && [[ "$TRAVIS_BRANCH" = "master" || -n "$TRAVIS_TAG" ]]; then
```
prydonius: we only want to run this on master and tags, and not pull-requests.

src/api/Makefile
```diff
@@ -42,7 +42,7 @@ swagger-serverstub:
 	rm -Rf ${SWAGGER_GEN_DIR}/cmd
 
 test:
-	${DEV_ENV_CMD} sh -c 'go test -v ./...'
+	${DEV_ENV_CMD} sh -c 'go test ./...'
```
prydonius: removed verbose mode because travis would fail due to too large a log

src/ui/src/app/repository-new/repository-new.component.html
```diff
@@ -10,11 +10,7 @@ <h2 class="repository-new__title">Add Repository</h2>
     </mat-form-field>
     <mat-form-field class="repository-new__form__field">
       <input matInput name="repository-url" placeholder="URL" required
-        [(ngModel)]="repoAttributes.URL" (ngModelChange)="formError=null">
-    </mat-form-field>
-    <mat-form-field class="repository-new__form__field">
-      <input matInput name="repository-source" placeholder="Source"
-        [(ngModel)]="repoAttributes.source" (ngModelChange)="formError=null">
+        [(ngModel)]="repoAttributes.url" (ngModelChange)="formError=null">
```
prydonius: the current `dev-v1` branch was failing due to the change of the repository model

.travis.yml
migmartri: And this is because your branch was triggering the build of the images?

.travis.yml
```diff
@@ -42,5 +42,6 @@ after_success:
 branches:
   only:
     - master
+    - dev-v1
```
migmartri: You might want to add a comment explaining that this is temporary and not to forget to remove it.

.travis.yml
prydonius: It's to prevent it, it wasn't triggering any builds at all before. The last time `latest` was pushed was 23 days ago, with the last commit to master.

.travis.yml
prydonius: Good point, thanks.

.travis.yml
```diff
@@ -1,6 +1,3 @@
-# Use older version of Trusty VM due to issues in the updated one
```
prydonius: this was a workaround before when we were having larger logs that was causing issues in newer travis VM versions, but now we've removed the verbose output I believe it is no longer needed

.travis.yml
```diff
@@ -18,11 +15,13 @@ before_install:
 install:
   - make -C src/ui install
   - make -C src/ui VERSION=$TRAVIS_TAG set-version
+  # required due to apparent yarn issue with node-sass: https://github.com/sass/node-sass/issues/1971
+  - npm --prefix src/ui rebuild node-sass
```
migmartri: Fun

Issue #505, opened 2018-09-11
Update documentation for Monocular 1.0


prydonius:
We should update the documentation to state:

- The main goal of the project to support the CNCF Helm Hub
- The set of features Monocular has, what environments it makes sense to use it in
- Remove in-cluster documentation and point to available alternatives.
- Updated developer documentation

PR #518, opened 2018-09-19
remove existing Monocular API implementation

prydonius:
We have replaced the main Monocular API server with chartsvc and the
chart-repo sync jobs as per
https://github.com/helm/community/blob/master/proposals/monocular-1.0-improvements.md.


prydonius:
@migmartri thanks, I added a note in https://github.com/helm/monocular/issues/505

PR #520, opened 2018-09-20
make foreground repository refresh as goroutine

fkpwolf:
To make pod respond to k8s liveness more quickly and get less pod restart times.


prydonius:
Thanks @fkpwolf, however with the changes in Monocular 1.0 to move chart syncing to Jobs/CronJobs I don't think this change is really needed. See https://github.com/helm/community/blob/master/proposals/monocular-1.0-improvements.md.

fkpwolf:
Thanks for the the information. The improvements looks nice.

prydonius:
Thanks for the feecback! FYI this work is going on in the dev-v1 branch https://github.com/helm/monocular/tree/dev-v1.

PR #523, opened 2018-09-21
change charts list component to extract repos from list of charts

prydonius:
The old Monocular API supported a repos endpoint that returned a list of
configured repos, but this endpoint is not supported by the chartsvc.
Instead of reimplementing it, we can extract the list of repos from the
charts list response client-side. This updates the chart list component
to always fetch all charts (regardless of a selected repo) allowing it
to extract the full list of repos to show. The full charts list is then
filtered by the selected repo to preserve the same functionality as the
/api/chartsvc/v1/charts/{repoName} endpoint previously used.

### DIFF
--- a/frontend/src/app/charts/charts.component.ts
+++ b/frontend/src/app/charts/charts.component.ts
@@ -1,8 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { ChartsService } from '../shared/services/charts.service';
-import { ReposService } from '../shared/services/repos.service';
 import { Chart } from '../shared/models/chart';
-import { Repo, RepoAttributes } from '../shared/models/repo';
+import { RepoAttributes } from '../shared/models/repo';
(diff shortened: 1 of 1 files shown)


frontend/src/app/charts/charts.component.ts
```diff
@@ -84,29 +77,36 @@ export class ChartsComponent implements OnInit {
...
-  loadRepos(): void {
-    this.reposService.getRepos().subscribe(repos => {
-      repos.splice(0, 0, this.allRepo);
-      this.filters[0].items = repos.map(r => ({
-        title: r.attributes.name,
-        value: r.id,
-        selected: this.repoName ? r.id == this.repoName : r.id == 'all'
-      }));
+  
+  setReposFromCharts(charts: Chart[]): void {
```
migmartri: It looks good. I'd just add a comment in the code explaining what this does and why it is needed.

Thanks!

PR #524, opened 2018-09-21
Remove in-cluster Monocular UI components

prydonius:
This commit removes frontend code related to installing and managing charts
and repositories in the app, when installed in an in-cluster environment.
Reasoning for this change is described in the Monocular 1.0 proposal
(https://github.com/helm/community/blob/master/proposals/monocular-1.0-improvements.md#removing-in-cluster-support).


frontend/src/app/app.routing.ts
```diff
@@ -5,42 +5,12 @@ import { PageNotFoundComponent } from './page-not-found/page-not-found.component
...
-    path: 'deployments/:deploymentName',
-    component: DeploymentComponent,
-    data: {
-      meta: {
-        title: 'Deployment information'
-      }
-    }
-  },
-  {
-    path: 'repos',
```
migmartri: How is repositories related to releases?

frontend/src/app/header-bar/header-bar.component.html
```diff
@@ -19,37 +19,12 @@ <h1>
             Charts
           </a>
         </li>
-        <li *ngIf="!loggedIn" class="separator"></li>
-        <li *ngIf="!loggedIn">
-          <a class="reverse" href="{{config.backendHostname}}/auth">
```
migmartri: And what about this?

Is this patch doing more than what the description says?

frontend/src/app/header-bar/header-bar.component.ts
```diff
@@ -6,7 +6,6 @@ import { DomSanitizer } from '@angular/platform-browser';
 import { MatIconRegistry } from '@angular/material';
 import { MatSnackBar } from '@angular/material';
 import { CookieService } from 'ngx-cookie';
-import { AuthService } from '../shared/services/auth.service';
```
migmartri: So removing the AuthService is part of the overall task?

frontend/src/app/app.routing.ts
prydonius: This page was added for an in-cluster feature to add and delete synced repos, it is not used in public site usecases such as https://hub.kubeapps.com. I believe this will not be that useful to the CNCF Helm Hub, but we can always revist it. The list of repos to browse can be seen on the charts page, e.g. https://hub.kubeapps.com/charts

frontend/src/app/header-bar/header-bar.component.html
prydonius: I updated the description to mention removal of the OAuth login. The chartsvc API doesn't support it, we may want to replace it with the LinuxFoundation identity login further down the line, but right now Monocular does not need it.

frontend/src/app/header-bar/header-bar.component.ts
prydonius: explained in https://github.com/helm/monocular/pull/524#discussion_r220023390

frontend/src/app/header-bar/header-bar.component.html
prydonius: After some thought, I've decided to keep the auth code in here, even though it is currently unused. This will make it easier for us to add LinuxFoundation login in the future.

prydonius:
Thanks for bringing it up @migmartri, I've updated this to re-add the auth code even if unused for now, it will be easier to support the LinuxFoundation login in the future.

migmartri:
Thanks for revisiting the changes :)

PR #525, opened 2018-09-22
update npm packages

prydonius:
attempt to address vulnerability issues

### DIFF
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -3,36 +3,36 @@
 
 
 "@angular-devkit/build-optimizer@~0.0.18":
-  version "0.0.21"
-  resolved "https://registry.yarnpkg.com/@angular-devkit/build-optimizer/-/build-optimizer-0.0.21.tgz#d3100a179c570d3c45ba02f87486c931e8845e13"
+  version "0.0.42"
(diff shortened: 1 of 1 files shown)


prydonius:
@migmartri mind taking a look at this one? I've run it locally and it seems to be fine, once we merge this into master we can see what vulnerability issues are still present.

prydonius:
@migmartri it's confusing because the reported vulnerabilities are coming from package-lock.json in master, and not yarn.lock. In the dev-v1 branch, package-lock.json has been removed and this PR is just a `yarn upgrade`. My thinking is to merge this and re-evaluate once dev-v1 is merged into master to see what packages are still vulnerable. What do you think?

migmartri:
> @migmartri it's confusing because the reported vulnerabilities are coming from package-lock.json in master, and not yarn.lock. In the dev-v1 branch, package-lock.json has been removed and this PR is just a `yarn upgrade`. My thinking is to merge this and re-evaluate once dev-v1 is merged into master to see what packages are still vulnerable. What do you think?

My comment was about the regular `package.json`, not the `.lock`. I agree with removing the `package.lock` file. 

I agree that this is a step in the right direction anyways.

PR #530, opened 2018-09-27
Improve the installation documentation

SilenceHVK:
In my reference to official Monocular installation method, there have been a lot of problems, here to improve the installation documentation.

- Need to install dns in k8s
- Need to install PV before installing Monocular

Or do not install PV by modifying values.yaml

```yaml
mongodb:
  mongodbDatabase: monocular
  persistence:
    enabled: false
```


prydonius:
Hey @SilenceHVK, we're getting close to release Monocular 1.0 (https://github.com/helm/monocular/tree/dev-v1), which disables persistence by default in MongoDB, so I don't think this change is really needed.

DNS in Kubernetes is generally expected by a lot of applications, I don't think it's really necessary to mention.

SilenceHVK:
Sounds good,thank you for your reply

PR #534, opened 2018-09-27
update travis config to push images to quay.io

prydonius:
Travis config in the UI has been updated to have DOCKER_USER and
DOCKER_PASSWORD point to a robot account in Quay.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ cache:
 services:
   - docker
 # UI dependencies
+env:
+  - IMAGE_PREFIX=quay.io/helmpack
 before_install:
   - export CHROME_BIN=chromium-browser
   - export DISPLAY=:99.0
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,6 +8,8 @@ cache:
 services:
   - docker
 # UI dependencies
+env:
+  - IMAGE_PREFIX=quay.io/helmpack
```
migmartri: Why do you need IMAGE_PREFIX to be a env variable? Does it change between runs?

prydonius:
> Does UI mean in the TravisCI project settings page?

yes, exactly!

.travis.yml
prydonius: because we'll later reuse it for chartsvc and chart-repo

PR #539, opened 2018-10-03
Switch to circleci

prydonius:
- based configuration off of https://github.com/kubeapps/kubeapps/blob/9a92cbf2bf0c618395389900d22567552282004f/.circleci/config.yml
- builds chartsvc and chart-repo
- template test for helm chart

This can be improved going forward, such as adding e2e tests before publishing images, but since those e2e tests do not exist today the building and pushing is combined into a single job to avoid the need to push intermediary images.

Some setup is required in the CircleCI config
- configure env vars for username and password
- configure github token for pushing to the repo to publish the chart

Once this is merged, TravisCI should be disabled.


migmartri:
Thanks,

This is useful.

prydonius:
@migmartri this is now ready to review

PR #547, opened 2018-10-24
External MongoDB chart option

### LINKED ISSUE #404: How do I use an external MongoDB instead?
I want to use an external database as a service MongoDB instance instead of the cluster one.  How would I do that?  Thanks.  

Also not sure why the MongoDB in the helm release is not starting up.  I get the following error: 

```
Welcome to the Bitnami mongodb container
Subscribe to project updates by watching https://github.com/bitnami/bitnami-docker-mongodb
Submit issues and feature requests at https://github.com/bitnami/bitnami-docker-mongodb/issues
Send us your feedback at containers@bitnami.com

nami    INFO  Initializing mongodb
Error executing 'postInstallation': Group '4294967294' not found
```

scottrigby:
@prydonius Here's a first pass at the external MongoDB chart option.

Not yet tested - but had some spare time tonight, so opening this for feedback.

The things I'm least sure about are chartsvc and chart-repo sync. I took a look at the [dbURL note](https://github.com/helm/monocular/blob/master/cmd/chartsvc/main.go#L60) about [mgo DIAL](https://godoc.org/labix.org/v2/mgo#Dial), and looked at [datastore.Config](https://github.com/kubeapps/common/blob/master/datastore/datastore.go#L28) briefly... it seems a fully formed `mongo-url` may work in place of the other commands (`mongo-user` with simple `mongo-url`) and `MONGO_PASSWORD` ENV var?

Once this PR is resolved, it should fix #404

### DIFF
--- a/chart/monocular/templates/chartsvc-deployment.yaml
+++ b/chart/monocular/templates/chartsvc-deployment.yaml
@@ -25,6 +25,9 @@ spec:
         command:
         - /chartsvc
         args:
+        {{- if and .Values.global.mongoUrl (not .Values.mongodb.enabled) -}}
+        - --mongo-url={{ .Values.global.mongoUrl }}
+        {{- else }}
         - --mongo-user=root
         - --mongo-url={{ template "mongodb.fullname" . }}
         env:
@@ -33,6 +36,7 @@ spec:
             secretKeyRef:
               name: {{ template "mongodb.fullname" . }}
               key: mongodb-root-password
+        {{- end }}
         ports:
         - name: http
           containerPort: {{ .Values.chartsvc.service.port }}
--- a/chart/monocular/templates/chartsvc-deployment.yaml
+++ b/chart/monocular/templates/chartsvc-deployment.yaml
@@ -25,7 +25,7 @@ spec:
         command:
         - /chartsvc
         args:
-        {{- if and .Values.global.mongoUrl (not .Values.mongodb.enabled) -}}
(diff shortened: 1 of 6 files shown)


prydonius:
Nice @scottrigby, yes I believe a full `mongo-url` can be used in place of the other flags and env var, so this should work. Were you able to try it out?

scottrigby:
I pinged @mattfarina on keybase, in case he has time to help me understand how to get the stable/monocular baseline set up before you get back 🇪🇸 ⚙️  

At least the checks are all passing now 😅

chart/monocular/templates/chartsvc-deployment.yaml
```diff
@@ -25,6 +25,9 @@ spec:
         command:
         - /chartsvc
         args:
+        {{- if and .Values.global.mongoUrl (not .Values.mongodb.enabled) -}}
```
mattfarina: `-}}` -> `}}`

mattfarina:
I've found another problem.

chartsvc isn't starting. If you use an external service you'll want to secure the connection with ssl. On the mongoUrl you'll have `?ssl=True` on the end. Unfortunately, this service uses `github.com/kubeapps/common` which is using `gopkg.in/mgo.v2`. `gopkg.in/mgo.v2` doesn't support that and is no longer maintained.

For example, because of this I can't seem to get it to connect to CosmosDB via the MongoDB API.

prydonius:
Let's fix this in kubeapps/common and then update that dependency in a separate PR, we can then release a new version of Monocular with the chart.

Issue #546, opened 2018-10-22
Getting "Unauthorized: not authorized on admin to execute command" in mongodb 


erihanse:
The chartsvc pod doesn't seem to be able to connect to the mongodb instance. When I look at the logs, there seems that the root user is not set up correctly. 
```
➜  monocular git:(master) ✗ kubectl logs monocular-release-mongodb-67dc9fbd58-j8rqv --namespace=erik-test --container=monocular-release-mongodb

Welcome to the Bitnami mongodb container
Subscribe to project updates by watching https://github.com/bitnami/bitnami-docker-mongodb
Submit issues and feature requests at https://github.com/bitnami/bitnami-docker-mongodb/issues

INFO  ==> Starting mongodb...
INFO  ==> Starting mongod...
2018-10-22T12:29:44.496+0000 I CONTROL  [main] Automatically disabling TLS 1.0, to force-enable TLS 1.0 specify --sslDisabledProtocols 'none'
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] MongoDB starting : pid=20 port=27017 dbpath=/opt/bitnami/mongodb/data/db 64-bit host=monocular-release-mongodb-67dc9fbd58-j8rqv
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] db version v4.1.4
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] git version: 2f4b5918497b09a226a3ec5dcff930edd52ea1e9
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] OpenSSL version: OpenSSL 1.0.1e-fips 11 Feb 2013
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] allocator: tcmalloc
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] modules: none
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten] build environment:
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten]     distmod: rhel70
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten]     distarch: x86_64
2018-10-22T12:29:44.504+0000 I CONTROL  [initandlisten]     target_arch: x86_64
[...]

PR #550, opened 2018-10-25
Update mongodb dependency

tompizmor:
It seems that updating MongoDB dependency version helped in this issue https://github.com/helm/monocular/issues/546


chart/monocular/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: monocular
 description: Monocular is a search and discovery front end for Helm Charts Repositories.
-version: 1.0.0
+version: 1.0.1
```
prydonius: can you make this 1.2.0 since mongodb is also a minor version update?

Issue #542, opened 2018-10-08
Cannot use monocular behind a proxy


jroucheton:
This issue for explaining that we cannot use monocular behind a proxy.

**WHY?**
Because, we cannot add a proxy though values.yaml
The new delivery (1.0) is based on jobs for synchronizing charts repositories.
A first one for the from scratch deployment and another one for the refresh mechanism.
Unfortunately, we cannot configure those jobs with dedicated proxy for downloading charts.

It will be nice to be able to add env vars (HTTP_PROXY, HTTPS_PROXY and NO_PROXY) at values level....

willuk2010:
@jroucheton check this out, 

https://github.com/kubeapps/kubeapps/issues/696

jroucheton:
Thanks @willuk2010 
Nevertheless, this proposal concerns kubeapps.
I hope that we can have the same kind of improvement for [Monocular](https://github.com/helm/monocular).

prydonius:
@jroucheton I think this just requires adding a new value to configure those env vars in the Jobs/CronJobs? Happy to accept a PR that does that!

PR #554, opened 2018-10-30
add http proxy variables for jobs

cvila84:
This is my proposal to fix https://github.com/helm/monocular/issues/542

Proxy now can be set in values.yaml, and will be passed as standard HTTP_PROXY and HTTPS_PROXY environment variables to chart-repo/sync which supports it as it is using standard GoLang http library


prydonius:
@cvila84 nice, thanks! We should add those values to the values.yaml file as well, commented out. Also, please bump the chart minor version!

You also need to sign-off on your commits, see https://github.com/helm/monocular/pull/554/checks

PR #558, opened 2018-11-09
Improve user agent and version handling in chart-repo command

migmartri:
* Adds an userAgent function that returns both the baked version and the application that is running it (i.e monocular), this application can be provided via a built-in flag.
* Adds a convenience version command
* Updates Makefile and circle-ci to pass the tagged release. By default unless specified, it will pass the commit sha.

```bash
$ make
docker run -it quay.io/helmpack/chart-repo:latest /chart-repo version
5d7825decbe8b1f1387b2ed916cb01013d214eac
$ make VERSION=1.0
docker run -it quay.io/helmpack/chart-repo:latest /chart-repo version
1.0
```

Next steps would be to update the Monocular chart to pass the new `application` flag.

Refs https://github.com/kubeapps/kubeapps/issues/767

### DIFF
--- a/cmd/chart-repo/chart_repo.go
+++ b/cmd/chart-repo/chart_repo.go
@@ -24,7 +24,7 @@ import (
 
 var rootCmd = &cobra.Command{
 	Use:   "chart-repo",
-	Short: "Kubeapps Chart Repository utility",
+	Short: "Monocular Chart Repository utility",
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
 	},
--- a/cmd/chart-repo/chart_repo.go
+++ b/cmd/chart-repo/chart_repo.go
@@ -24,7 +24,7 @@ import (
 
 var rootCmd = &cobra.Command{
 	Use:   "chart-repo",
-	Short: "Monocular Chart Repository utility",
+	Short: "Chart Repository utility",
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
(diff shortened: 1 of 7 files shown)


cmd/chart-repo/chart_repo.go
```diff
@@ -24,7 +24,7 @@ import (
 
 var rootCmd = &cobra.Command{
 	Use:   "chart-repo",
-	Short: "Kubeapps Chart Repository utility",
+	Short: "Monocular Chart Repository utility",
```
prydonius: Perhaps we can just call this "Chart Repository utility" since it is a standalone tool used in multiple projects.

cmd/chart-repo/chart_repo.go
```diff
@@ -45,6 +45,9 @@ func init() {
 		cmd.Flags().String("mongo-url", "localhost", "MongoDB URL (see https://godoc.org/labix.org/v2/mgo#Dial for format)")
 		cmd.Flags().String("mongo-database", "charts", "MongoDB database")
 		cmd.Flags().String("mongo-user", "", "MongoDB user")
+		// see version.go
+		cmd.Flags().StringVarP(&application, "application", "", "", "Name of the application using chart-repo")
```
prydonius: I would suggest being more explicit about what this is for, e.g. `user-agent-comment` or `user-agent-bundle` and explain in the description that it can be used to provide more info in the user agent string when querying chart repository servers.

PR #568, opened 2018-11-22
Add option to set MongoDB URL string in a secret

scottrigby:
On one hand this could be further simplified by removing the string option, and only allowing the secret option. But that would require a MAJOR version bump (removing an existing option), whereas this only requires MINOR. @prydonius What do you think?

----
Simplify the conditional logic with "mongodb.url" helper definition.

Also remove HTTP_PROXY AND HTTPS_PROXY ENV vars outside the mongodb.enabled
conditional wrapper - I believe that was an unintentional result of two PRs
merging cleanly in close proximity:
- https://github.com/helm/monocular/pull/547
- https://github.com/helm/monocular/pull/554

Signed-off-by: Scott Rigby <scott@r6by.com>

### DIFF
--- a/chart/monocular/templates/_helpers.tpl
+++ b/chart/monocular/templates/_helpers.tpl
@@ -31,6 +31,18 @@ Render image reference
 {{- end -}}
 
 {{/*
+MongoDB URL argument.
+*/}}
+{{- define "mongodb.url" -}}
+{{- if $global.Values.mongodb.enabled }}
+{{ template "mongodb.fullname" $global }}
+{{- else if $global.Values.global.mongoUrlSecret }}
+{{ "$MONGO_URL" }}
+{{- else if $global.Values.global.mongoUrl }}
+{{ $global.Values.global.mongoUrl }}
+{{- end -}}
+
+{{/*
 Sync job pod template
 */}}
 {{- define "monocular.sync.podTemplate" -}}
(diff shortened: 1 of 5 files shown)


chart/monocular/templates/_helpers.tpl
```diff
@@ -49,28 +49,37 @@ spec:
...
-    {{- if and $global.Values.global.mongoUrl (not $global.Values.mongodb.enabled) }}
-    - --mongo-url={{ $global.Values.global.mongoUrl }}
-    {{- else }}
+    {{- if $global.Values.mongodb.enabled }}
     - --mongo-url={{ template "mongodb.fullname" $global }}
     - --mongo-user=root
+    {{- else if $global.Values.global.mongoUrl }}
+    - --mongo-url={{ $global.Values.global.mongoUrl }}
+    {{- else if $global.Values.global.mongoUrlSecret }}
+    - --mongo-url={{ "$MONGO_URL" }}
```
prydonius: Are the `{{ }}` needed here? I think this could just be

```suggestion
    - --mongo-url="$MONGO_URL"
```

chart/monocular/templates/_helpers.tpl
```diff
@@ -49,28 +49,37 @@ spec:
...
     - name: HTTPS_PROXY
       value: {{ $global.Values.sync.httpsProxy }}
+    {{- if $global.Values.mongodb.enabled }}
     - name: MONGO_PASSWORD
       valueFrom:
         secretKeyRef:
           key: mongodb-root-password
           name: {{ template "mongodb.fullname" $global }}
     {{- end }}
+    {{- if $global.Values.global.mongoUrlSecret }}
```
prydonius: this could be elseif I guess, though does not really matter much...

chart/monocular/values.yaml
```diff
@@ -148,5 +148,10 @@ mongodb:
 # This must be set if mongodb.enabled is set to false, following the pattern:
 # `mongodb://${MONGODB_USER}:${MONGODB_ROOT_PASSWORD}@${MONGODB_DNS}:${MONGODB_PORT}/${MONGODB_DATABASE}`
 # ref: https://docs.mongodb.com/manual/reference/connection-string/
+# You may set the connection URL in one of two ways:
+# - mongoUrl: store the connection string directly in Helm values
+# - mongoUrlSecret: name of the secret where the connection string is stored,
+#   where the key is "mongo-url-secret". See NOTES.txt.
 global:
```
prydonius: What do you think about moving this configuration to `externalDatabase`? e.g. https://github.com/helm/charts/blob/master/stable/wordpress/values.yaml#L78. Then in the future we are one step closer to splitting up the options in host, user, password, etc. It would make this a breaking change, but I doubt anyone is using this setting yet so the risk should be very low.

Issue #570, opened 2018-12-05
Deal with duplicate digests from different repositories in chartsvc


prydonius:
Some chart repositories might contain duplicate charts that exist in other repositories. One example of this is the Bitnami charts repository, which syncs several charts from the Helm stable repository. In Monocular, when both repositories are added, multiple entries will be seen in the UI for the same charts (though listed as being from different repositories). This comes across as confusing, and I think a better UX would be to simply merge the entries, picking the first repository found.

We could make a change in the chartsvc `listCharts` handler:
https://github.com/helm/monocular/blob/master/cmd/chartsvc/handler.go#L65
to filter the list of charts and remove charts where the latest version has a digest we have seen before.

One additional issue related to this is that the filtering of charts from specific repositories in the charts list happens client-side (https://github.com/helm/monocular/blob/master/frontend/src/app/charts/charts.component.ts#L87), so we would need to change this to actually request from the chartsvc endpoint for the repository (https://github.com/helm/monocular/blob/master/cmd/chartsvc/handler.go#L80) to ensure we don't miss out ignored charts in the full charts list response.

We should implement this in two steps, first changing the `listCharts` handler as described, and then updating the frontend to request charts by repo when set as a separate PR.

PR #571, opened 2018-12-05
Filter list response with unique results

andresmgot:

Ref: #570

Return unique charts by digest when listing.

cc/ @prydonius

### DIFF
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -217,15 +217,14 @@ func newChartResponse(c *models.Chart) *apiResponse {
 }
 
 func newChartListResponse(charts []*models.Chart) apiListResponse {
-	chartDict := map[string]*apiResponse{}
+	// We will keep track of unique digest:chart to avoid duplicates
+	chartDigests := map[string]bool{}
 	cl := apiListResponse{}
 	for _, c := range charts {
-		chartResponse := newChartResponse(c)
-		// The same chart may be present in different repos, filter just one
-		id := (chartResponse.Relationships["latestChartVersion"].Data).(models.ChartVersion).Digest
-		if found := chartDict[id]; found == nil {
-			chartDict[id] = chartResponse
-			cl = append(cl, chartResponse)
+		digest := c.ChartVersions[0].Digest
+		if _, ok := chartDigests[digest]; !ok {
+			chartDigests[digest] = true
+			cl = append(cl, newChartResponse(c))
 		}
 	}
 	return cl
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -217,11 +217,17 @@ func newChartResponse(c *models.Chart) *apiResponse {
 }
 
 func newChartListResponse(charts []*models.Chart) apiListResponse {
+	chartDict := map[string]*apiResponse{}
 	cl := apiListResponse{}
 	for _, c := range charts {
-		cl = append(cl, newChartResponse(c))
+		chartResponse := newChartResponse(c)
+		// The same chart may be present in different repos, filter just one
+		id := (chartResponse.Relationships["latestChartVersion"].Data).(models.ChartVersion).Digest
+		if found := chartDict[id]; found == nil {
+			chartDict[id] = chartResponse
+			cl = append(cl, chartResponse)
+		}
 	}
-
 	return cl
 }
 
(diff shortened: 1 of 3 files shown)


cmd/chartsvc/handler.go
```diff
@@ -217,11 +217,17 @@ func newChartResponse(c *models.Chart) *apiResponse {
 }
 
 func newChartListResponse(charts []*models.Chart) apiListResponse {
+	chartDict := map[string]*apiResponse{}
```
prydonius: I would call this `chartDigests` and add a comment above this to describe that this is used to track chart versions by digests to ensure uniqueness

cmd/chartsvc/handler.go
```diff
@@ -217,11 +217,17 @@ func newChartResponse(c *models.Chart) *apiResponse {
...
 	cl := apiListResponse{}
 	for _, c := range charts {
-		cl = append(cl, newChartResponse(c))
+		chartResponse := newChartResponse(c)
+		// The same chart may be present in different repos, filter just one
+		id := (chartResponse.Relationships["latestChartVersion"].Data).(models.ChartVersion).Digest
```
prydonius: this might be confusing because the `id` in the response and mongodb is `reponame/chartname`, not the digest, we could just call this variable `digest`

cmd/chartsvc/handler.go
```diff
@@ -217,11 +217,17 @@ func newChartResponse(c *models.Chart) *apiResponse {
...
+		if found := chartDict[id]; found == nil {
```
prydonius: whilst this does work, I think a more correct way of doing this would be:

```
if _, ok := chartDict[id]; !ok {
...
}
```

The use of `ok` instead of `found` is a Go idiom that I see regularly, so I think it's better to use that name.

cmd/chartsvc/handler.go
prydonius: wouldn't it be simpler to just do

```suggestion
		id := c.ChartVersions[0].Digest
```
That way we avoid having to assert the type.

We can use `newChartResponse` below when we append to `cl`.

cmd/chartsvc/handler_test.go
```diff
@@ -126,28 +126,40 @@ func Test_newChartResponse(t *testing.T) {
...
+			{ID: "stable/wordpress", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "1234"}, {Version: "1.2.2", Digest: "12345"}}},
+		}, []*models.Chart{
+			{ID: "my-repo/my-chart", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
+			{ID: "stable/wordpress", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "1234"}, {Version: "1.2.2", Digest: "12345"}}},
+		}},
+		{"has a duplicated chart", []*models.Chart{
+			{ID: "my-repo/my-chart", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
+			{ID: "other-repo/my-chart", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
+		}, []*models.Chart{
+			{ID: "my-repo/my-chart", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
```
prydonius: I think one other case here is if two different charts have an older version that is duplicated, but the latest versions are different (might be a strange edge case), but I think the expected thing to do in this case would be to not filter it out. wdyt? worth adding that test?

cmd/chartsvc/handler.go
```diff
@@ -217,11 +217,16 @@ func newChartResponse(c *models.Chart) *apiResponse {
 }
 
 func newChartListResponse(charts []*models.Chart) apiListResponse {
+	// We will keep track of unique digest:chart to avoid duplicates
+	chartDigests := map[string]bool{}
 	cl := apiListResponse{}
 	for _, c := range charts {
-		cl = append(cl, newChartResponse(c))
+		digest := c.ChartVersions[0].Digest
+		if _, ok := chartDigests[digest]; !ok {
```
prydonius: I still think it's useful here to add a comment like "filter out the chart if we've seen the same digest before" just as an added clarification to the comment above

cmd/chartsvc/handler_test.go
```diff
@@ -126,28 +126,47 @@ func Test_newChartResponse(t *testing.T) {
...
+		}},
+		{"has a duplicated chart but with different versions", []*models.Chart{
```
prydonius:
```suggestion
		{"has a duplicated older chart version but different latest versions", []*models.Chart{
```

cmd/chartsvc/handler.go
andresmgot:
```suggestion
		// Filter out the chart if we've seen the same digest before
		if _, ok := chartDigests[digest]; !ok {
```

PR #572, opened 2018-12-05
fetch list of repos to show in ui from server

prydonius:
This adds a new repos.json file that holds the information about
what repositories are being synced, allowing us to remove the current
hack that parses the full charts list response for the list of
repositories. With this, we can once again use the /charts/{repo}
endpoint to fetch charts from a single repo.

closes #572

### DIFF
--- a/frontend/src/app/shared/services/repos.service.ts
+++ b/frontend/src/app/shared/services/repos.service.ts
@@ -38,5 +38,4 @@ export class ReposService {
     console.error(errMsg); // log to console instead
     return Observable.throw(errMsg);
   }
-
 }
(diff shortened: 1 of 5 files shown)


frontend/src/app/shared/services/repos.service.ts
```diff
@@ -0,0 +1,42 @@
...
+    return body.data || { };
+  }
+
+  private handleError (error: any) {
+    let errMsg = (error.json().message) ? error.json().message :
+      error.status ? `${error.status} - ${error.statusText}` : 'Server error';
+    console.error(errMsg); // log to console instead
+    return Observable.throw(errMsg);
+  }
+
```
migmartri: extra line

PR #575, opened 2018-12-06
Footer updates

mattfarina:
screen shot 2018-12-06 at 3 49 58 pm


mattfarina:
Note, the language on the CNCF and incubating is taken directly from helm.sh.

PR #576, opened 2018-12-09
add a link to tell people how to submit a repo

michelleN:


frontend/src/app/footer/footer.component.html
```diff
@@ -7,7 +7,7 @@
   <app-footer-list [title]="'Contribute'">
     <li><a class="reverse" target="_blank" href="https://github.com/helm/hub">Helm Hub</a></li>
     <li><a class="reverse" target="_blank" href="https://github.com/helm/monocular">Monocular</a></li>
-    <li><a class="reverse" target="_blank" href="https://github.com/helm/helm">Helm</a></li>
```
prydonius: any reason to remove this one? We can probably just add this as an additional link? although I suppose it might look weird having 4 vs 3 in this column?

PR #578, opened 2018-12-13
Allow to use a custom CA bundle in chart-repo

andresmgot:
Reference: https://github.com/kubeapps/kubeapps/issues/515

Allow users to mount a custom CA certificate under `/etc/registry-ca` (hardcoded for the moment) used when syncing/fetching registries.

cc/ @prydonius

### DIFF
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -20,12 +20,15 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"path"
 	"strings"
 	"sync"
@@ -44,6 +47,8 @@ const (
 	chartCollection       = "charts"
 	chartFilesCollection  = "files"
 	defaultTimeoutSeconds = 10
+	caRegistryDir         = "/etc/registry-ca"
+	rootCARegistry        = "/etc/ssl/certs/ca-certificates.crt"
 )
 
 type importChartFilesJob struct {
@@ -56,13 +61,19 @@ type httpClient interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
-var netClient httpClient = &http.Client{
-	Timeout: time.Second * defaultTimeoutSeconds,
-}
+var netClient httpClient = &http.Client{}
 
 func parseRepoUrl(repoURL string) (*url.URL, error) {
 	repoURL = strings.TrimSpace(repoURL)
 	return url.ParseRequestURI(repoURL)
+}
+
+func init() {
+	var err error
+	netClient, _, err = initNetClient(rootCARegistry, caRegistryDir)
+	if err != nil {
+		log.Fatal(err)
+	}
 }
 
 // Syncing is performed in the following steps:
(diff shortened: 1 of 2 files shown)


cmd/chart-repo/utils.go
```diff
@@ -44,6 +47,8 @@ const (
 	chartCollection       = "charts"
 	chartFilesCollection  = "files"
 	defaultTimeoutSeconds = 10
+	caRegistryDir         = "/etc/registry-ca"
```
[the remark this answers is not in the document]
andresmgot: I have set it as a directory for two reasons:

 - To be able to use several CA files. It's possible to append several CA certificates together but it's also possible to use different files.
 - To remove the assumption of what will be the name of the file. People can name that file in different ways: `ca.crt`, `ca.pem`, `ca.crt.pem`... If we read the directory we don't need to force a specific name.

cmd/chart-repo/utils.go
```diff
@@ -44,6 +47,8 @@ const (
...
+	rootCARegistry        = "/etc/ssl/certs/ca-certificates.crt"
```
migmartri: I was confused by this. I did not know that this one is the path to the system ca-certificates. I thought that it was the custom rootCA that you want to use.

I'd rename it to systemCA or something like that and mention that you need it since it is required once appending CAs

cmd/chart-repo/utils.go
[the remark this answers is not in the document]
migmartri: I see, I personally think that it is unnecessary, I mean, I am not sure how common is to use multiple CA certificates, and as you said they can be chained in a single file.

w.r.t the name, I do not think that there is nothing wrong with enforcing a specific name (Tiller does it for example https://github.com/helm/helm/blob/master/cmd/tiller/tiller.go#L272) actually we control in the chart how it is going to be mounted. 

If we focus on only supporting a single file the code will get simplified.

cmd/chart-repo/utils.go
migmartri: I gave a quick look at Tiller's implementation and it seems that they do not require the systemCAs but instead create an empty pool? https://github.com/helm/helm/blob/7cad59091a9451b2aa4f95aa882ea27e6b195f98/pkg/tlsutil/tls.go#L75

cmd/chart-repo/utils.go
```diff
@@ -415,3 +426,45 @@ func chartTarballURL(r repo, cv chartVersion) string {
 	}
 	return source
 }
+
+func initNetClient(rootCACert, additionalCACertDir string) (*http.Client, *http.Transport, error) {
+	caCertPool := x509.NewCertPool()
+	systemCACert, err := ioutil.ReadFile(rootCACert)
```
prydonius: Can't we just load the default pool using `x509.SystemCertPool()`, falling back to `NewCertPool` if that comes back empty? See https://forfuncsake.github.io/post/2017/08/trust-extra-ca-cert-in-go-app/?

cmd/chart-repo/utils.go
```diff
@@ -415,3 +426,45 @@ func chartTarballURL(r repo, cv chartVersion) string {
...
+				caCert, err := ioutil.ReadFile(fullPath)
+				if err != nil {
+					return nil, nil, fmt.Errorf("Unable to read CA cert file: %v", err)
+				}
+				ok := caCertPool.AppendCertsFromPEM(caCert)
+				log.Printf("Adding CA cert from %s: %v", fullPath, ok)
+			}
+		}
+	}
+	// Return Transport for testing purposes
```
prydonius: can't we just access the transport using `client.Transport`?

andresmgot: In any case I have changed the method and the test to avoid that and just check for errors.

cmd/chart-repo/utils.go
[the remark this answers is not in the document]
prydonius: @andresmgot good to know, thanks

cmd/chart-repo/utils.go
```diff
@@ -44,6 +47,7 @@ const (
 	chartCollection       = "charts"
 	chartFilesCollection  = "files"
 	defaultTimeoutSeconds = 10
+	additionalCAFile      = "/usr/local/share/ca-certificates/ca.crt"
```
prydonius: 👍

PR #579, opened 2018-12-17
Use '<<' instead of '<<<'

choldrim:
I don't know if '<<<' is right, but '<<<' doesn't work for me whatever zsh or bash


prydonius:
@choldrim are you able to sign off your commit using the instructions here: https://github.com/helm/monocular/pull/579/checks?

choldrim:
Sorry to reply that so late, I have committed that patch with signed-off option, please check if it's ok.

PR #586, opened 2019-01-03
Add security context

obeyler:
This PR allows to set a securityContext to be able to launch monocular in a Kubernetes with a PodSecurityContext set as runAsNonRoot


prydonius:
Happy new year @obeyler!!

PR #592, opened 2019-01-15
Add /resolve-repo endpoint to the chartsvc

andresmgot:
Ref: https://github.com/kubeapps/kubeapps/issues/588

This PR adds a new endpoint to the chart-svc `/resolve-repo`. This endpoint will receive as a POST some chart metadata (that is contained in the release information) and will return the list of repositories that contains that chart. It also includes the latest version available in that repository.

cc/ @migmartri @prydonius

### DIFF
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -200,33 +200,26 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
-func findLatestChart(name, version, appversion string) []*models.Chart {
+// listChartsWithFilters returns the list of repos that contains the given chart and the latest version found
+func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Params) {
 	db, closer := dbSession.DB()
 	defer closer()
 
 	var charts []*models.Chart
 	if err := db.C(chartCollection).Find(bson.M{
-		"name": name,
+		"name": params["chartName"],
 		"chartversions": bson.M{
-			"$elemMatch": bson.M{"version": version, "appversion": appversion},
-		}}).All(&charts); err != nil {
-		log.WithError(err).Errorf("could not find the given chart with id %s", name)
+			"$elemMatch": bson.M{"version": req.FormValue("version"), "appversion": req.FormValue("appversion")},
+		}}).Select(bson.M{
+		"name": 1, "repo": 1,
+		"chartversions": bson.M{"$slice": 1},
+	}).All(&charts); err != nil {
+		log.WithError(err).Errorf("could not find the given chart with id %s", params["chartName"])
 		// continue to return empty list
 	}
 
-	return charts
-}
-
-// resolveRepos returns the list of repos that contains the given chart and the latest version found
-func resolveRepos(w http.ResponseWriter, req *http.Request, params Params) {
-	latests := findLatestChart(params["chartName"], req.FormValue("version"), req.FormValue("appversion"))
-
-	var cvl apiListResponse
-	for _, cv := range latests {
-		cvl = append(cvl, newChartResponse(cv))
-	}
-
-	response.NewDataResponse(cvl).Write(w)
+	cl := newChartListResponse(charts)
+	response.NewDataResponse(cl).Write(w)
 }
 
 func newChartResponse(c *models.Chart) *apiResponse {
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -214,7 +214,10 @@ func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Para
 		"name": 1, "repo": 1,
 		"chartversions": bson.M{"$slice": 1},
 	}).All(&charts); err != nil {
-		log.WithError(err).Errorf("could not find the given chart with id %s", params["chartName"])
+		log.WithError(err).Errorf(
+			"could not find charts with the given name %s, version %s and appversion %s",
+			params["chartName"], req.FormValue("version"), req.FormValue("appversion"),
+		)
 		// continue to return empty list
 	}
 
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -17,18 +17,14 @@ limitations under the License.
 package main
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
-	"reflect"
 
 	"github.com/globalsign/mgo/bson"
 	"github.com/gorilla/mux"
 	"github.com/helm/monocular/cmd/chartsvc/models"
 	"github.com/kubeapps/common/response"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/helm/pkg/proto/hapi/chart"
 )
 
 // Params a key-value map of path params
@@ -204,88 +200,35 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
-func isChartContained(chartData *models.Chart, chartMetadata *chart.Metadata) bool {
-	eq := true
-	// Compare simple fields
-	if chartData.Description != chartMetadata.Description ||
-		chartData.Home != chartMetadata.Home ||
-		chartData.Icon != chartMetadata.Icon {
-		eq = false
-	}
-	// Compare array of key words (but it may be empty)
-	if len(chartData.Keywords) > 0 && len(chartMetadata.Keywords) > 0 {
-		if !reflect.DeepEqual(chartData.Keywords, chartMetadata.Keywords) {
-			eq = false
-		}
-	}
-	// Compare array of maintainers (but it may be empty)
-	if len(chartData.Maintainers) > 0 && len(chartMetadata.Maintainers) > 0 {
-		// Maintainers format is different (pointers versus structs)
-		c2MaintainerArray := []chart.Maintainer{}
-		for _, m := range chartMetadata.Maintainers {
-			c2MaintainerArray = append(c2MaintainerArray, *m)
-		}
-		if !reflect.DeepEqual(chartData.Maintainers, c2MaintainerArray) {
-			eq = false
-		}
-	}
-	if eq {
-		// Charts matches, look if the version exists
-		for _, v := range chartData.ChartVersions {
-			if v.Version == chartMetadata.Version && v.AppVersion == chartMetadata.AppVersion {
-				return true
-			}
-		}
-	}
-	return false
-}
-
-// resolveRepo returns the repo that contains the given chart and the latest version found
-func resolveRepo(w http.ResponseWriter, req *http.Request) {
+func findLatestChart(name, version, appversion string) []models.ChartLatest {
 	db, closer := dbSession.DB()
 	defer closer()
 
-	metadataBytes, err := ioutil.ReadAll(req.Body)
-	if err != nil {
-		errStr := fmt.Sprintf("could not read body %v", err)
-		log.WithError(err).Errorf(errStr)
-		response.NewErrorResponse(http.StatusUnprocessableEntity, errStr).Write(w)
-		return
-	}
-	chartMetadata := &chart.Metadata{}
-	err = json.Unmarshal(metadataBytes, chartMetadata)
-	if err != nil {
-		errStr := fmt.Sprintf("could not read body %v", err)
-		log.WithError(err).Errorf(errStr)
-		response.NewErrorResponse(http.StatusUnprocessableEntity, errStr).Write(w)
-		return
-	}
-
 	var charts []*models.Chart
-	if err := db.C(chartCollection).Find(bson.M{"name": chartMetadata.Name}).All(&charts); err != nil {
-		log.WithError(err).Errorf("could not find the given chart with id %s", chartMetadata.Name)
+	if err := db.C(chartCollection).Find(bson.M{
+		"name": name,
+		"chartversions": bson.M{
+			"$elemMatch": bson.M{"version": version, "appversion": appversion},
+		}}).All(&charts); err != nil {
+		log.WithError(err).Errorf("could not find the given chart with id %s", name)
 		// continue to return empty list
 	}
 
 	latest := []models.ChartLatest{}
 	for _, c := range charts {
-		if isChartContained(c, chartMetadata) {
-			// We rely that versions are stored in order (newer first) to return the latest
-			latest = append(latest, models.ChartLatest{
-				Chart:  c.Name,
-				Latest: c.ChartVersions[0].Version,
-				Repo:   c.Repo.Name,
-			})
-		}
+		// We rely that versions are stored in order (newer first) to return the latest
+		latest = append(latest, models.ChartLatest{
+			Name:           c.Name,
+			LatestVersion:  c.ChartVersions[0].Version,
+			RepositoryName: c.Repo.Name,
+		})
 	}
+	return latest
+}
 
-	latestRaw, err := json.Marshal(latest)
-	if err != nil {
-		log.WithError(err).Errorf("could not render result %v", err)
-		response.NewErrorResponse(http.StatusInternalServerError, "could render result").Write(w)
-		return
-	}
-	w.Write(latestRaw)
+// resolveRepos returns the list of repos that contains the given chart and the latest version found
+func resolveRepos(w http.ResponseWriter, req *http.Request, params Params) {
+	response.NewDataResponse(findLatestChart(params["chartName"], req.FormValue("version"), req.FormValue("appversion"))).Write(w)
 }
 
 func newChartResponse(c *models.Chart) *apiResponse {
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -200,7 +200,7 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
-func findLatestChart(name, version, appversion string) []models.ChartLatest {
+func findLatestChart(name, version, appversion string) []*models.Chart {
 	db, closer := dbSession.DB()
 	defer closer()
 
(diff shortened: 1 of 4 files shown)


cmd/chartsvc/main.go
```diff
@@ -47,6 +47,7 @@ func setupRoutes() http.Handler {
 	apiv1.Methods("GET").Path("/charts/{repo}/{chartName}").Handler(WithParams(getChart))
 	apiv1.Methods("GET").Path("/charts/{repo}/{chartName}/versions").Handler(WithParams(listChartVersions))
 	apiv1.Methods("GET").Path("/charts/{repo}/{chartName}/versions/{version}").Handler(WithParams(getChartVersion))
+	apiv1.Methods("POST").Path("/charts/resolve-repo").HandlerFunc(resolveRepo)
```
migmartri: I am not sure that POST is the correct verb for this endpoint. To me, POST is associated with a Non idempotent action (creation usually). In this case we have an idempotent action that returns a resource so it makes sense to me to use GET instead.

That said, I am assuming that you used POST to send the data in the body. I am wondering if we should send the manifest as base64encoded via query params instead.

`/charts/resolve-repo?manifest=[base64Manifest]`

Note that using Get with query params help with server side caching.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+	if err != nil {
+		errStr := fmt.Sprintf("could not read body %v", err)
+		log.WithError(err).Errorf(errStr)
+		response.NewErrorResponse(http.StatusUnprocessableEntity, errStr).Write(w)
+		return
+	}
+	chartMetadata := &chart.Metadata{}
+	err = json.Unmarshal(metadataBytes, chartMetadata)
+	if err != nil {
+		errStr := fmt.Sprintf("could not read body %v", err)
```
migmartri: This error message might not be specific enough since it is the same one than before.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+		log.WithError(err).Errorf(errStr)
+		response.NewErrorResponse(http.StatusUnprocessableEntity, errStr).Write(w)
+		return
+	}
+
+	var charts []*models.Chart
+	if err := db.C(chartCollection).Find(bson.M{"name": chartMetadata.Name}).All(&charts); err != nil {
```
migmartri: It would help to me to see some comments explaining the procedure.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
+func isChartContained(chartData *models.Chart, chartMetadata *chart.Metadata) bool {
+	eq := true
```
migmartri: maybe we should start with false?

cmd/chartsvc/models/chart.go
```diff
@@ -62,3 +60,10 @@ type ChartFiles struct {
 	Readme string
 	Values string
 }
+
+// ChartLatest express the latest version for a chart in a repo
+type ChartLatest struct {
+	Chart  string `json:"chart"`
+	Latest string `json:"latest"`
+	Repo   string `json:"repo"`
```
migmartri: Couldn't we have their proper types in this struct? It's difficult to know in this case which one is a Chart, is it `Chart`? Is it `Latest`?

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+	// Compare simple fields
+	if chartData.Description != chartMetadata.Description ||
+		chartData.Home != chartMetadata.Home ||
+		chartData.Icon != chartMetadata.Icon {
+		eq = false
```
migmartri: At this point you should have returned no?

cmd/chartsvc/models/chart.go
migmartri: Update:

I see now, I'd rename those fields to make them more explicit.

Chart => Name
Latest => LatestVersion
Repo => RepositoryName

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+			// We rely that versions are stored in order (newer first) to return the latest
+			latest = append(latest, models.ChartLatest{
+				Chart:  c.Name,
+				Latest: c.ChartVersions[0].Version,
+				Repo:   c.Repo.Name,
+			})
+		}
+	}
+
+	latestRaw, err := json.Marshal(latest)
```
migmartri: Do you need to do this marshaling? In other handlers we seem to do something like `response.NewDataResponse(latest).Write(w)`, which also seems to add the `data` key as described in the https://jsonapi.org/ spec. Any reason for doing it differently here?

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
+func isChartContained(chartData *models.Chart, chartMetadata *chart.Metadata) bool {
```
migmartri: Is chart contained or maybe a better name would be isSameChart? That's what you are checking no?

cmd/chartsvc/main.go
migmartri: I am not either convinced with the name of the endpoint either. This endpoint will return the latest versions of the chart in every single repo that it belongs to no?

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+			c2MaintainerArray = append(c2MaintainerArray, *m)
+		}
+		if !reflect.DeepEqual(chartData.Maintainers, c2MaintainerArray) {
+			eq = false
+		}
+	}
+	if eq {
+		// Charts matches, look if the version exists
+		for _, v := range chartData.ChartVersions {
+			if v.Version == chartMetadata.Version && v.AppVersion == chartMetadata.AppVersion {
```
migmartri: I am a little bit confused, so this code returns if there is a chart in the repo that maps the one that we are asking about with exactly the same version. Does that mean that we are not returning the actual latest?

cmd/chartsvc/handler.go
```diff
@@ -200,6 +204,90 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// resolveRepo returns the repo that contains the given chart and the latest version found
```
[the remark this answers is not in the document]
prydonius: actually we should rename this and the endpoint `resolveRepos` given we'll potentially get back more than one.

cmd/chartsvc/main.go
prydonius: +1 I think we should try to use GET if we can. We'd need to check if there are any restrictions here on URL length though.

cmd/chartsvc/models/chart.go
```diff
@@ -62,3 +60,10 @@ type ChartFiles struct {
 	Readme string
 	Values string
 }
+
+// ChartLatest express the latest version for a chart in a repo
+type ChartLatest struct {
```
prydonius: the name of this struct is a bit strange, maybe it's better if we call this `ResolveRepoResponse`. Also I'm wondering if we should try to follow the `apiResponse` structure (https://github.com/helm/monocular/blob/e607f09567c23a6d0401facfc77eb9df0d4aeaf6/cmd/chartsvc/handler.go#L48) to keep the API more consistent. We could use `apiListResponse` with each ID being the name of the repo and attributes for the chartName and latestVersion. WDYT?

cmd/chartsvc/models/chart.go
andresmgot: okay

andresmgot: okay

cmd/chartsvc/main.go
andresmgot: at the end I've changed it because `/charts/resolve-repo/{chartName}` conflicts with `/charts/{repo}/{chartName}`. I think a neat solution would be `GET /repos/resolve/{chartName}?version={version}&appversion={appversion}`

cmd/chartsvc/handler.go
andresmgot: do you mean explaining the mongo query?

cmd/chartsvc/models/chart.go
andresmgot: I am not sure if this comment still applies?

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,37 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
+func findLatestChart(name, version, appversion string) []models.ChartLatest {
```
prydonius: What's the reason for the separate function? The `params` parameter in the `resolveRepos` handler does make the handler testable.

migmartri:
> Continued from https://github.com/helm/monocular/pull/592/files#r248130628.
> 
> I think we should try to use the apiResponse structure (derived from the JSON API spec: https://jsonapi.org/), since we are using it everywhere else. Is there a specific reason we don't want to use it for this?
> 
> I also think we need a bit more thought about the endpoint we want here and what we expect to be returned. My feeling is that what we want to return here are primarily Repository resources, with some relationship about the latest version of a particular Chart in the Repository. This is based on your suggestion of the endpoint being `/repos/resolve/?name={chartName}&version={version}&appversion={appversion}` (though note that I don't think the subresource should be the name of the chart, but rather a query parameter). In this case, I would structure the response in the following way:
> 
> ```
> [
>   {
>     ID: "stable"
>     Type: "repository"
>     Attributes: { Name: "stable", URL: "..." }
>     Relationships: {
>       latestChartVersion: {...}
>     }
>   },
>   ...
> ]
> ```
> Here we can reuse `latestChartVersion` from the getChart response (
> 
> [monocular/cmd/chartsvc/handler.go](https://github.com/helm/monocular/blob/e607f09567c23a6d0401facfc77eb9df0d4aeaf6/cmd/chartsvc/handler.go#L298-L301)
> 
> Lines 298 to 301 in [e607f09](/helm/monocular/commit/e607f09567c23a6d0401facfc77eb9df0d4aeaf6)
> 
>  Relationships: relMap{ 
>  	"latestChartVersion": rel{ 
>  		Data:  chartVersionAttributes(c.ID, latestCV), 
>  		Links: selfLink{pathPrefix + "/charts/" + c.ID + "/versions/" + latestCV.Version}, 
> ), but we may want to adapt it to include the Chart name too.
> Alternatively, if we think that we really need a specific resource type for this feature, the endpoint should reflect it. For example, it might make sense to have the endpoint be `/resolve/{chartName}?version={version}&appversion={appversion}` (in other words, resolve the chart of name `chartName` with these extra parameters), and we could return a "ResolvedChart" type. It could be structured like this:
> 
> ```
> {
>   ID: "wordpress"
>   Type: "resolvedChart"
>   Attributes: {
>     Name: "wordpress"
>     LatestVersions: [
>       { RepositoryName: "stable", Version: "1.0.0" },
>       ...
>     ]
>   }
> }
> ```
> Note, the above is just a proposal. I'm not sure what the best option is, but I think overall I prefer the Repository resource response with relationship, as it feels more reusable. It's difficult to say though as this is a bit of a specific usecase. My main concern is that we make an informed decision about the endpoint and response type.

Either option work for me although I've a slight subconscious preference for the later. Mainly because since I do not see this new resource to fit into any existing one, the fact that will be mounted outside of any known path is a relieve for me, it also delegates the decision to later (naming is hard :P)

cmd/chartsvc/handler.go
migmartri: Not really, it was more about explaining the whole body of the method. Something like, it will fetch foo, then do bar and finally filter by zed.

andresmgot:
> I think we should try to use the apiResponse structure (derived from the JSON API spec: https://jsonapi.org/), since we are using it everywhere else. Is there a specific reason we don't want to use it for this?

Sorry, you are right, I missed that standard in the code

> I also think we need a bit more thought about the endpoint we want here and what we expect to be returned

What do you think about using `/resolve/{chartName}?version={version}&appversion={appversion}` as the endpoint and the ResolvedChart type as:

```
	return &apiResponse{
		Type:       "resolvedChart",
		ID:         c.ID,
		Attributes: chartAttributes(*c),
		Links:      selfLink{pathPrefix + "/charts/" + c.ID},
		Relationships: relMap{
			"latestChartVersion": rel{
				Data:  chartVersionAttributes(c.ID, latestCV),
				Links: selfLink{pathPrefix + "/charts/" + c.ID + "/versions/" + latestCV.Version},
			},
		},
	}
```

That's is very similar to the response for `/charts/{chartName}/versions/{version}` being `c` the chartVersion that's requested in the URL and latestCV the latest one. This contains all the info for the Dashboard.

prydonius:
As discussed in the call, I like the idea of returning Chart resources filtered by name. Given that it's now the same as the `/charts` response, we can have the endpoint be `/charts?name={chartName}&version={version}&appVersion={appVersion}`. It'll be a little bit odd since you won't actually get the ChartVersion object back for the requested version and appVersion, but I think that is okay as long as we add a comment explaining the usecase and in the future we can always add an extra relationship for the requested ChartVersion.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,37 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+	db, closer := dbSession.DB()
+	defer closer()
+
+	var charts []*models.Chart
+	if err := db.C(chartCollection).Find(bson.M{
+		"name": name,
+		"chartversions": bson.M{
+			"$elemMatch": bson.M{"version": version, "appversion": appversion},
+		}}).All(&charts); err != nil {
```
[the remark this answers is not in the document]
andresmgot: Actually now we need to return all the chartversions to find the latest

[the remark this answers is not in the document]
prydonius: Perhaps I'm missing something, but we want all the charts, but only the latest chartversion (which will always be the first element in the `chartversions` array). The example query I shared earlier does exactly this:


```
db.charts.find({"name": "wordpress", "chartversions": {"$elemMatch": {"version": "5.0.4", "appversion": "5.0.2"}}}, {"repo": 1, "name": 1, "chartversions": { "$slice": 1 }})
```

```
{ "_id" : "bitnami/wordpress", "name" : "wordpress", "repo" : { "name" : "bitnami", "url" : "https://charts.bitnami.com/bitnami" }, "chartversions" : [ { "version" : "5.1.1", "appversion" : "5.0.3", "created" : ISODate("2019-01-15T13:40:37.478Z"), "digest" : "eb4867b9802356b60b58d6f11bfc87a9194440975eeec7601cd4fe161e70af64", "urls" : [ "https://charts.bitnami.com/bitnami/wordpress-5.1.1.tgz" ] } ] }
{ "_id" : "stable/wordpress", "name" : "wordpress", "repo" : { "name" : "stable", "url" : "https://kubernetes-charts.storage.googleapis.com" }, "chartversions" : [ { "version" : "5.1.2", "appversion" : "5.0.3", "created" : ISODate("2019-01-16T00:25:35.630Z"), "digest" : "3fb906523d6079b10438ab0b60a5c7e3dccde8c2ab51fb67e1a2f4cf57ba169b", "urls" : [ "https://kubernetes-charts.storage.googleapis.com/wordpress-5.1.2.tgz" ] } ] }
```

Isn't that what we want?

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,35 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
 	w.Write([]byte(files.Values))
 }
 
+func findLatestChart(name, version, appversion string) []*models.Chart {
```
prydonius: This isn't really finding the latest chart more than just finding the chart with filters, right? I believe this/or the handler should be called `listChartsWithFilters` or similar. Although I'm still not sure we need a separate function for this here, given we mock the HTTP requests everywhere else.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,35 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+		}}).All(&charts); err != nil {
+		log.WithError(err).Errorf("could not find the given chart with id %s", name)
+		// continue to return empty list
+	}
+
+	return charts
+}
+
+// resolveRepos returns the list of repos that contains the given chart and the latest version found
+func resolveRepos(w http.ResponseWriter, req *http.Request, params Params) {
```
prydonius: as suggested above, I think this handler should be `listChartsWithFilters` or something similar.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,35 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+	latests := findLatestChart(params["chartName"], req.FormValue("version"), req.FormValue("appversion"))
+
+	var cvl apiListResponse
```
prydonius: These aren't cv (chartversion) objects, they're chart objects which contain cv objects. I would not use the `cv` shorthand here. The `listCharts` method uses the shorthand `cl`: https://github.com/helm/monocular/pull/592/files#diff-b8a9743d32f40304acc271b43066ee95R75

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,35 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+	for _, cv := range latests {
```
prydonius: Are we not using `newChartListResponse` because in this case we want duplicates? If so, can we add a comment above this for-loop to mention this?

cmd/chartsvc/handler.go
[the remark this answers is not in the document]
andresmgot: I tried that query and it was not returning the latest chart in the `chartversions` array but the chart version that matches `version:appversion`. Let me double check to see if was my error translating the syntax to Golang or something like that.

cmd/chartsvc/handler.go
andresmgot: I still think that having atomic functions that don't deal with different things is a slightly better approach but since the rest of methods follow that approach I am okay merging the methods.

cmd/chartsvc/handler.go
andresmgot: Nope, I didn't notice that method.

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,28 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+	var charts []*models.Chart
+	if err := db.C(chartCollection).Find(bson.M{
+		"name": params["chartName"],
+		"chartversions": bson.M{
+			"$elemMatch": bson.M{"version": req.FormValue("version"), "appversion": req.FormValue("appversion")},
+		}}).Select(bson.M{
+		"name": 1, "repo": 1,
+		"chartversions": bson.M{"$slice": 1},
+	}).All(&charts); err != nil {
+		log.WithError(err).Errorf("could not find the given chart with id %s", params["chartName"])
```
prydonius: I think this error message was copied and pasted, it should probably just say something generic like "could not find charts with filters"

cmd/chartsvc/handler.go
```diff
@@ -200,6 +200,28 @@ func getChartVersionValues(w http.ResponseWriter, req *http.Request, params Para
...
+		// continue to return empty list
+	}
+
+	cl := newChartListResponse(charts)
```
prydonius: This helper will filter out repos if the latest version is a duplicate from a different repo. I think this is okay, because if the `bitnami` and `stable` repo have the same latest versions, it'll just show one of them. However, if one of them had a newer version that the other, it will return both since the latest versions are not duplicates.

Just double checking since you're more aware with how this fits in with the other changes and requirements, does this work the way we want?

andresmgot: Yes, I am aware of that behavior. I think it's okay to return just one chart (from a single repo) in that case since we are just reporting that the version exists.

andresmgot:
@prydonius @migmartri I still don't have write permissions in this repository, should I have them?

PR #597, opened 2019-01-31
make repo schedule and successful Job history count configurable

cpoole:
Added two variables to make the job history count configurable

### DIFF
--- a/chart/monocular/templates/repo-sync-cronjobs.yaml
+++ b/chart/monocular/templates/repo-sync-cronjobs.yaml
@@ -19,7 +19,8 @@ spec:
     spec:
       template:
 {{ include "monocular.sync.podTemplate" (list . $) | indent 8 }}
-  schedule: "0 * * * *"
-  successfulJobsHistoryLimit: 3
+  schedule: {{ default "0 * * * *" .schedule | quote }}
+  successfulJobsHistoryLimit: {{ default 3 .successfulJobsHistoryLimit }}
+ 
 ---
 {{- end -}}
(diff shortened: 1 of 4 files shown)


chart/monocular/templates/repo-sync-cronjobs.yaml
```diff
@@ -19,7 +19,16 @@ spec:
     spec:
       template:
 {{ include "monocular.sync.podTemplate" (list . $) | indent 8 }}
-  schedule: "0 * * * *"
-  successfulJobsHistoryLimit: 3
+  schedule: {{ if not .schedule -}}
```
prydonius: I think this can be simplified to a one-liner with `{{ default "0 * * * *" .schedule | quote }}`

chart/monocular/templates/repo-sync-cronjobs.yaml
```diff
@@ -19,7 +19,16 @@ spec:
...
+    "0 * * * *"
+  {{- else -}}
+    {{ .schedule | quote }}
+  {{- end }}
+  successfulJobsHistoryLimit: {{ if not .successfulJobsHistoryLimit -}}
```
prydonius: Likewise, this could be made simpler with default.

chart/monocular/templates/repo-sync-cronjobs.yaml
cpoole: wow, this is awesome. I did not know sprig had a default function.

cpoole:
updated to address feedback @prydonius

cpoole:
I bumped to 1.2.9 but im happy to bump to 1.3.0 if you'd prefer

prydonius:
Np! It's a new backwards-compatible feature (new values) in the chart, so should be a minor bump as far as semver goes.

cpoole:
rebased back onto upstream. 

also friendly bump :slightly_smiling_face:

prydonius:
sorry missed this - thanks!

PR #598, opened 2019-02-13
chart: Add networkPolicy

pho-enix:
Add networkPolicy for services. Needed to run monocular in environments where default policy is deny.


prydonius:
thanks!

Issue #603, opened 2019-02-21
chart-repo: panics when there are no charts to import


prydonius:
First reported on Slack (https://kubernetes.slack.com/archives/C9D3TSUG4/p1550784319000900), it looks like if https://github.com/helm/monocular/blob/v1.2.0/cmd/chart-repo/utils.go#L100 comes back with an empty list (if there are no entries in the index.yaml, or all charts are deprecated), then https://github.com/helm/monocular/blob/v1.2.0/cmd/chart-repo/utils.go#L260 results in a panic as it tries to access the first element of the `charts` array.

We should exit earlier with a suitable error message when `charts` is empty.

Dean-Coakley:
@prydonius This can be closed?

PR #604, opened 2019-02-22
Added check for empty charts in chart-repo

hpandeycodeit:
Added check for empty charts : Issue #603

### DIFF
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -98,6 +98,10 @@ func syncRepo(dbSession datastore.Session, repoName, repoURL string, authorizati
 	}
 
 	charts := chartsFromIndex(index, r)
+	if len(charts) == 0 {
+		log.Fatal("No charts are found")
+		return nil
+	}
 	err = importCharts(dbSession, charts)
 	if err != nil {
 		return err
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -99,7 +99,7 @@ func syncRepo(dbSession datastore.Session, repoName, repoURL string, authorizati
 
 	charts := chartsFromIndex(index, r)
 	if len(charts) == 0 {
-		log.Fatal("No charts are found")
+		errors.New("no charts in repository index")
 		return nil
 	}
 	err = importCharts(dbSession, charts)
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -99,8 +99,7 @@ func syncRepo(dbSession datastore.Session, repoName, repoURL string, authorizati
 
 	charts := chartsFromIndex(index, r)
 	if len(charts) == 0 {
-		errors.New("no charts in repository index")
-		return nil
+		return errors.New("no charts in repository index")
 	}
 	err = importCharts(dbSession, charts)
 	if err != nil {
--- a/cmd/chart-repo/utils_test.go
+++ b/cmd/chart-repo/utils_test.go
@@ -73,6 +73,24 @@ func (h *goodHTTPClient) Do(req *http.Request) (*http.Response, error) {
 	return w.Result(), nil
 }
 
+type chartHTTPClient struct{}
+
+func (h *chartHTTPClient) Do(req *http.Request) (*http.Response, error){
+	w := httptest.NewRecorder()
+	// Don't accept trailing slashes
+	if strings.HasPrefix(req.URL.Path, "//") {
+		w.WriteHeader(500)
+	}
+	// If subpath repo URL test, check that index.yaml is correctly added to the
+	// subpath
+	if req.URL.Host == "subpath.test" && req.URL.Path != "/subpath/index.yaml" {
+		w.WriteHeader(500)
+	}
+
+	w.Write(nil)
+	return w.Result(), nil
+}
+
 type authenticatedHTTPClient struct{}
 
 func (h *authenticatedHTTPClient) Do(req *http.Request) (*http.Response, error) {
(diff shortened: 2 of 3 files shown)


cmd/chart-repo/utils.go
```diff
@@ -98,6 +98,10 @@ func syncRepo(dbSession datastore.Session, repoName, repoURL string, authorizati
 	}
 
 	charts := chartsFromIndex(index, r)
+	if len(charts) == 0 {
+		log.Fatal("No charts are found")
```
prydonius: can I suggest simply returning an `errors.New("no charts in repository index")`, as the caller of this function will log the error with Fatal: https://github.com/helm/monocular/blob/e19c1b1818e0e611257db71247f7d08705ac42f2/cmd/chart-repo/sync.go#L64.

The issue with using Fatal here is that it will exit the program here, and that should really be the job of the caller of this function. This function should just return a response and not control the flow of the program.

hpandeycodeit: @prydonius Thanks for explaining. 

Do you mean to add a test something like this: 
``` 
type chartHTTPClient struct{}

func (h *chartHTTPClient) Do(req *http.Request) (*http.Response, error){
	w := httptest.NewRecorder()
	// Don't accept trailing slashes
	if strings.HasPrefix(req.URL.Path, "//") {
		w.WriteHeader(500)
	}
	// If subpath repo URL test, check that index.yaml is correctly added to the
	// subpath
	if req.URL.Host == "subpath.test" && req.URL.Path != "/subpath/index.yaml" {
		w.WriteHeader(500)
	}
	w.Write(nil)
	return w.Result(), nil
}

Comment (prydonius) [reply cmd/chart-repo/utils.go]: @hpandeycodeit I think we can remove the trailing slashes and subpath parts in that, and just return an empty index. We should also call this something like `emptyChartRepoHTTPClient` to make the intent more clear.

I'm also not sure if `w.Write(nil)` will be sufficient though, we might need to create a test index file (like https://github.com/helm/monocular/blob/e19c1b1818e0e611257db71247f7d08705ac42f2/cmd/chart-repo/testdata/valid-index.yaml#L1) that has an `entries` key that is empty.

e.g.:

```
type emptyChartRepoHTTPClient struct{}

func (h *emptyChartRepoHTTPClient) Do(req *http.Request) (*http.Response, error){
	w := httptest.NewRecorder()

        w.Write([]byte(emptyRepoIndexYAML))
	return w.Result(), nil
}
```

We would then need a test that is similar to https://github.com/helm/monocular/blob/e19c1b1818e0e611257db71247f7d08705ac42f2/cmd/chart-repo/utils_test.go#L157, but also sets the `netClient` like https://github.com/helm/monocular/blob/e19c1b1818e0e611257db71247f7d08705ac42f2/cmd/chart-repo/utils_test.go#L202, and asserts than an error is returned.

Does that make sense?

Comment (hpandeycodeit) [reply cmd/chart-repo/utils.go]: @prydonius Yes, it does. I have added the test case. Let me know if it looks okay to you. 

:)

File: cmd/chart-repo/utils.go [hunk 2]
```diff
@@ -98,6 +98,10 @@ func syncRepo(dbSession datastore.Session, repoName, repoURL string, authorizati
 	}
 
 	charts := chartsFromIndex(index, r)
+	if len(charts) == 0 {
+		errors.New("no charts in repository index")
```
prydonius: We'll need to return the error so that the caller can exit appropriately

cmd/chart-repo/utils_test.go
```diff
@@ -73,6 +73,24 @@ func (h *goodHTTPClient) Do(req *http.Request) (*http.Response, error) {
 	return w.Result(), nil
 }
 
+type chartHTTPClient struct{}
```
prydonius: I guess we can remove this one and the `Do` below now that we have the emptyChartRepoHTTPClient?

cmd/chart-repo/utils_test.go
```diff
@@ -570,3 +588,22 @@ h251U/Daz6NiQBM9AxyAw6EHm8XAZBvCuebfzyrT
...
+func (h *emptyChartRepoHTTPClient) Do(req *http.Request) (*http.Response, error){
+	w := httptest.NewRecorder()
+	w.Write([]byte(emptyRepoIndexYAML))
+	return w.Result(), nil
+}  
+
+
+func Test_emptyChartRepo(t *testing.T) {
+	netClient = &badHTTPClient{} 
+	_, err := fetchRepoIndex(repo{URL: "https://my.examplerepo.com"})
```
prydonius: we actually need to call `syncRepo`, as this is where the error for an empty repo is found

cmd/chart-repo/utils_test.go
```diff
@@ -570,3 +588,22 @@ h251U/Daz6NiQBM9AxyAw6EHm8XAZBvCuebfzyrT
...
+
+func (h *emptyChartRepoHTTPClient) Do(req *http.Request) (*http.Response, error){
+	w := httptest.NewRecorder()
+	w.Write([]byte(emptyRepoIndexYAML))
+	return w.Result(), nil
+}  
+
+
+func Test_emptyChartRepo(t *testing.T) {
+	netClient = &badHTTPClient{}
```
prydonius:
```suggestion
	netClient = &emptyChartRepoHTTPClient{} 
```

hpandeycodeit: done

cmd/chart-repo/utils_test.go
hpandeycodeit: done

cmd/chart-repo/utils_test.go
hpandeycodeit: Looks like I forgot to remove this one. :) 
Removed this one now.

cmd/chart-repo/utils.go
hpandeycodeit: done

prydonius:
Thanks @hpandeycodeit!

hpandeycodeit:
Thank you @prydonius!

PR #605, opened 2019-02-27
chartsvc: Add support for pagination

andresmgot:

This PR adds support for requesting a certain number of charts (and in consequence, divide the response in pages) for the endpoints `/charts` and `/charts/repo`. 

In case the new parameters (`page` and `size`) are not given, the default behavior is maintained (return all the charts in a single page) to avoid breaking changes.

cc/ @prydonius

### DIFF
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -18,11 +18,14 @@ package main
 
 import (
 	"fmt"
+	"math"
 	"net/http"
+	"strconv"
 
 	"github.com/globalsign/mgo/bson"
 	"github.com/gorilla/mux"
 	"github.com/helm/monocular/cmd/chartsvc/models"
+	"github.com/kubeapps/common/datastore"
 	"github.com/kubeapps/common/response"
 	log "github.com/sirupsen/logrus"
 )
@@ -61,34 +64,97 @@ type rel struct {
 	Links selfLink    `json:"links"`
 }
 
-// listCharts returns a list of charts
-func listCharts(w http.ResponseWriter, req *http.Request) {
+type meta struct {
+	TotalPages int `json:"totalPages"`
+}
+
+// getPageNumberAndSize extracts the page number and size of a request. Default (1, 0) if not set
+func getPageNumberAndSize(req *http.Request) (int, int) {
+	pageNumber := req.FormValue("page")
+	pageSize := req.FormValue("size")
+	numberInt, err := strconv.Atoi(pageNumber)
+	if err != nil || numberInt < 1 {
+		numberInt = 1
+	}
+	sizeInt, _ := strconv.Atoi(pageSize)
+	// Safeguard for negative numbers
+	if sizeInt < 0 {
+		sizeInt = 0
+	}
+	return numberInt, sizeInt
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func getChartList(repo string, pageNumber, pageSize int) (apiListResponse, interface{}, error) {
 	db, closer := dbSession.DB()
 	defer closer()
 	var charts []*models.Chart
-	if err := db.C(chartCollection).Find(nil).Sort("name").All(&charts); err != nil {
+
+	var query datastore.Query
+	c := db.C(chartCollection)
+	if repo != "" {
+		query = c.Find(bson.M{"repo.name": repo})
+	} else {
+		query = c.Find(nil)
+	}
+	query = query.Sort("name")
+
+	// TODO(andresmgot): We should not query the charts just yet in case we paginate the response
+	// but we need to do so since it's not possible to filter & count charts (countDocuments) until
+	// https://github.com/kubeapps/common/issues/8 is addressed
+	err := query.All(&charts)
+	if err != nil {
+		return apiListResponse{}, 0, err
+	}
+
+	totalPages := 1
+	if pageSize != 0 {
+		// If a pageSize is given, returns only the the specified number of charts and
+		// the number of pages
+		length := len(charts)
+		totalPages = int(math.Ceil(float64(length) / float64(pageSize)))
+
+		// If the page number is out of range, return the last one
+		if pageNumber > totalPages {
+			pageNumber = totalPages
+		}
+
+		first := pageSize * (pageNumber - 1)
+		last := min(first+pageSize, length)
+		charts = charts[first:last]
+	}
+
+	return newChartListResponse(charts), meta{totalPages}, nil
+}
+
+// listCharts returns a list of charts
+func listCharts(w http.ResponseWriter, req *http.Request) {
+	pageNumber, pageSize := getPageNumberAndSize(req)
+	cl, meta, err := getChartList("", pageNumber, pageSize)
+	if err != nil {
 		log.WithError(err).Error("could not fetch charts")
 		response.NewErrorResponse(http.StatusInternalServerError, "could not fetch all charts").Write(w)
 		return
 	}
-
-	cl := newChartListResponse(charts)
-	response.NewDataResponse(cl).Write(w)
+	response.NewDataResponseWithMeta(cl, meta).Write(w)
 }
 
 // listRepoCharts returns a list of charts in the given repo
 func listRepoCharts(w http.ResponseWriter, req *http.Request, params Params) {
-	db, closer := dbSession.DB()
-	defer closer()
-	var charts []*models.Chart
-	if err := db.C(chartCollection).Find(bson.M{"repo.name": params["repo"]}).Sort("_id").All(&charts); err != nil {
+	pageNumber, pageSize := getPageNumberAndSize(req)
+	cl, meta, err := getChartList(params["repo"], pageNumber, pageSize)
+	if err != nil {
 		log.WithError(err).Error("could not fetch charts")
 		response.NewErrorResponse(http.StatusInternalServerError, "could not fetch all charts").Write(w)
 		return
 	}
-
-	cl := newChartListResponse(charts)
-	response.NewDataResponse(cl).Write(w)
+	response.NewDataResponseWithMeta(cl, meta).Write(w)
 }
 
 // getChart returns the chart from the given repo
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -67,8 +67,9 @@ type meta struct {
 	TotalPages int `json:"totalPages"`
 }
 
+// count is used to parse the result of a $count operation in the database
 type count struct {
-	Count int `json:"count"`
+	Count int
 }
 
 // getPageNumberAndSize extracts the page number and size of a request. Default (1, 0) if not set
--- a/cmd/chartsvc/handler.go
+++ b/cmd/chartsvc/handler.go
@@ -70,18 +70,14 @@ type meta struct {
 
 // getPageNumberAndSize extracts the page number and size of a request. Default (1, 0) if not set
 func getPageNumberAndSize(req *http.Request) (int, int) {
-	pageNumber := req.FormValue("page")
-	pageSize := req.FormValue("size")
-	numberInt, err := strconv.Atoi(pageNumber)
-	if err != nil || numberInt < 1 {
-		numberInt = 1
+	page := req.FormValue("page")
+	size := req.FormValue("size")
+	pageInt, err := strconv.ParseUint(page, 10, 64)
+	if err != nil {
+		pageInt = 1
 	}
-	sizeInt, _ := strconv.Atoi(pageSize)
-	// Safeguard for negative numbers
-	if sizeInt < 0 {
-		sizeInt = 0
-	}
-	return numberInt, sizeInt
+	sizeInt, _ := strconv.ParseUint(size, 10, 64)
+	return int(pageInt), int(sizeInt)
 }
 
 func min(a, b int) int {
@@ -89,6 +85,21 @@ func min(a, b int) int {
 		return a
 	}
 	return b
+}
+
+func uniqChartList(charts []*models.Chart) []*models.Chart {
+	// We will keep track of unique digest:chart to avoid duplicates
+	chartDigests := map[string]bool{}
+	res := []*models.Chart{}
+	for _, c := range charts {
+		digest := c.ChartVersions[0].Digest
+		// Filter out the chart if we've seen the same digest before
+		if _, ok := chartDigests[digest]; !ok {
+			chartDigests[digest] = true
+			res = append(res, c)
+		}
+	}
+	return res
 }
 
 func getChartList(repo string, pageNumber, pageSize int) (apiListResponse, interface{}, error) {
@@ -112,6 +123,8 @@ func getChartList(repo string, pageNumber, pageSize int) (apiListResponse, inter
 	if err != nil {
 		return apiListResponse{}, 0, err
 	}
+	// Filter duplicated charts
+	charts = uniqChartList(charts)
 
 	totalPages := 1
 	if pageSize != 0 {
@@ -287,7 +300,7 @@ func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Para
 		// continue to return empty list
 	}
 
-	cl := newChartListResponse(charts)
+	cl := newChartListResponse(uniqChartList(charts))
 	response.NewDataResponse(cl).Write(w)
 }
 
@@ -308,16 +321,9 @@ func newChartResponse(c *models.Chart) *apiResponse {
 }
 
 func newChartListResponse(charts []*models.Chart) apiListResponse {
-	// We will keep track of unique digest:chart to avoid duplicates
-	chartDigests := map[string]bool{}
 	cl := apiListResponse{}
 	for _, c := range charts {
-		digest := c.ChartVersions[0].Digest
-		// Filter out the chart if we've seen the same digest before
-		if _, ok := chartDigests[digest]; !ok {
-			chartDigests[digest] = true
-			cl = append(cl, newChartResponse(c))
-		}
+		cl = append(cl, newChartResponse(c))
 	}
 	return cl
 }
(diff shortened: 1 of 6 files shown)


cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,97 @@ type rel struct {
...
+	c := db.C(chartCollection)
+	if repo != "" {
+		query = c.Find(bson.M{"repo.name": repo})
+	} else {
+		query = c.Find(nil)
+	}
+	query = query.Sort("name")
+
+	// TODO(andresmgot): We should not query the charts just yet in case we paginate the response
+	// but we need to do so since it's not possible to filter & count charts (countDocuments) until
```
prydonius: I don't understand, why can't we use `count`? Some quick testing shows it works:

```
> db.charts.find().count()
345
> db.charts.find({"repo.name": "stable"}).count()
238
> db.charts.find({"repo.name": "bitnami"}).count()
54
```

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,97 @@ type rel struct {
...
+		if pageNumber > totalPages {
+			pageNumber = totalPages
+		}
+
+		first := pageSize * (pageNumber - 1)
+		last := min(first+pageSize, length)
+		charts = charts[first:last]
+	}
+
+	return newChartListResponse(charts), meta{totalPages}, nil
```
prydonius: How do we plan to handle the de-duplication of charts from different repos? `newChartListResponse` will de-dupe by digest, but if not all charts are loaded, we won't be able to fully de-dupe. This might be okay because we will at least de-dupe the charts that are shown on the page, but charts between pages will be shown, but we should look into what this looks like and add a comment considering it.

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,97 @@ type rel struct {
...
+		length := len(charts)
+		totalPages = int(math.Ceil(float64(length) / float64(pageSize)))
+
+		// If the page number is out of range, return the last one
+		if pageNumber > totalPages {
+			pageNumber = totalPages
+		}
+
+		first := pageSize * (pageNumber - 1)
+		last := min(first+pageSize, length)
```
prydonius: `math.Min` https://golang.org/pkg/math/#Min?

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,97 @@ type rel struct {
...
+
+// getPageNumberAndSize extracts the page number and size of a request. Default (1, 0) if not set
+func getPageNumberAndSize(req *http.Request) (int, int) {
+	pageNumber := req.FormValue("page")
+	pageSize := req.FormValue("size")
+	numberInt, err := strconv.Atoi(pageNumber)
+	if err != nil || numberInt < 1 {
+		numberInt = 1
+	}
+	sizeInt, _ := strconv.Atoi(pageSize)
```
prydonius: I would use https://golang.org/pkg/strconv/#ParseUint which returns an error if `sizeInt < 0`

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,97 @@ type rel struct {
...
-func listCharts(w http.ResponseWriter, req *http.Request) {
+type meta struct {
+	TotalPages int `json:"totalPages"`
+}
+
+// getPageNumberAndSize extracts the page number and size of a request. Default (1, 0) if not set
+func getPageNumberAndSize(req *http.Request) (int, int) {
+	pageNumber := req.FormValue("page")
+	pageSize := req.FormValue("size")
+	numberInt, err := strconv.Atoi(pageNumber)
```
prydonius:
```suggestion
	pageNumberInt, err := strconv.Atoi(pageNumber)
```

or simply rename pageNumber -> page everywhere

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,97 @@ type rel struct {
...
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func getChartList(repo string, pageNumber, pageSize int) (apiListResponse, interface{}, error) {
```
prydonius: `getPaginatedChartList`?

cmd/chartsvc/handler_test.go
```diff
@@ -229,16 +230,81 @@ func Test_newChartVersionListResponse(t *testing.T) {
...
+			{ID: "my-repo/my-chart", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
+			{ID: "stable/dokuwiki", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "1234"}}},
+			{ID: "stable/drupal", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "12345"}}},
+			{ID: "stable/wordpress", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "123456"}}},
+		}, []*models.Chart{
+			{ID: "my-repo/my-chart", ChartVersions: []models.ChartVersion{{Version: "0.0.1", Digest: "123"}}},
+			{ID: "stable/dokuwiki", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "1234"}}},
+			{ID: "stable/drupal", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "12345"}}},
+			{ID: "stable/wordpress", ChartVersions: []models.ChartVersion{{Version: "1.2.3", Digest: "123456"}}},
+		}, meta{1}},
```
prydonius: we should add tests about the expected behaviour when de-duplicating

cmd/chartsvc/handler.go
andresmgot: Yes, that's kind of what we are doing here. My point is that when doing `db.charts.find()` you are querying the whole collection while, if using pagination, you are only interested in a certain number of charts:

```
db.charts.find().skip(5).limit(5)
```

That would query just the second page of 5 elements.

ref: https://www.codementor.io/arpitbhayani/fast-and-efficient-pagination-in-mongodb-9095flbqr

Now the goal is to obtain the total amount of pages without querying all the documents in the collection. For doing so, I assume that `db.charts.countDocuments({"repo.name": "stable")}` is faster since it doesn't return the documents but just the sum of elements.

In any case we are not working with huge collections so the optimizations are minimal.

cmd/chartsvc/handler.go
andresmgot: That's a good point, I missed it. The issue is that, if there is a duplicated chart we will be returning `size-1` charts which will be weird. I think it's better to filter unique charts before this point so we can properly calculate the amount of pages shown and return the proper number of charts.

cmd/chartsvc/handler.go
prydonius: Okay, I think I understand what your concern is, but I think you might be misunderstanding how this works.

Firstly, `db.C(chartCollection).Find(...)` only prepares a query, it will not execute it[0]. It gets executed when the returned `Query` Object is called with `All`, `One`, `Count` and other methods that cause an execution.

Secondly, `count` can be optimised in certain cases. As you mentioned, when there isn't a query, it will be performed in constant time[1] by pulling from the collections stats. When there is a query, though, it can be slower. However, indexes in the database would make this faster. I'm not 100% sure about the difference between `countDocuments` and `count`, but from what I can find `countDocuments` is supposedly more accurate in sharded clusters due to using the aggregation API. Also, `count` is apparently deprecated in Mongo v4, but I still think it's worth using it here (we should probably add a comment).

One way to see that `count` is handled a little differently is that it ignores the `skip` and `limit` filters when counting (this is a similar behaviour to `SELECT COUNT(*)` in SQL):

```
> db.charts.find().skip(5).limit(5).count()
345
```

You can use `db.charts.explain()...` to compare how MongoDB handles different queries. Suffice to say, I think we can more than reasonably use `count` to get the total count and then perform the query with skip and limit for retrieving the actual resources. Another very good reason for this is that we avoid loading all charts in memory when we only care about a portion.

> In any case we are not working with huge collections so the optimizations are minimal.

Agreed, which is why we shouldn't need to worry too much about the performance impact of a `count` operation, but fetching everything and doing pagination client-side is an anti-pattern. Databases are designed for this sort of thing, and will be much better at optimising it.

[0] https://godoc.org/github.com/globalsign/mgo#Collection.Find
[1] https://stackoverflow.com/questions/49984134/performance-of-count-for-mongodb

cmd/chartsvc/handler.go
prydonius: This makes using `count` as above difficult. We should spend some time looking into ways to craft a MongoDB query to filter unique charts.

cmd/chartsvc/handler.go
prydonius: Can you add a comment above your `min` function to explain that?

cmd/chartsvc/handler.go
prydonius: Just a thought, I might be confused as to why `listCharts` doesn't implement this directly. `getPaginatedChartList` tells me that there is something more to this.

cmd/chartsvc/handler.go
prydonius: Okay, I was able to craft a query using the aggregation API that is able to return charts with distinct digests:

```
db.charts.aggregate(
  [
    {
      $addFields:
        {
          firstChartVersion: { $arrayElemAt: [ "$chartversions", 0 ] },
        }
    },
    {
      $group: {
        _id: "$firstChartVersion.digest",
        chart: { $first: "$$ROOT" }
      }
    },
    {
      $replaceRoot: {
        newRoot: "$chart"
      }
    }
  ]
)
```

Explanation:
1. addFields - This copies the first element of the chartversions array to a new field, firstChartVersion. This is necessary for the next pipeline, as I couldn't find a way of grouping by a value in the first object in an array.
2. group - This groups all charts by the digest in the latest chartversion, and so removes any duplicates. Because this will return a new document with `_id` set to the digest, this also adds a new field `chart` that contains the original object so we can recover it.
3. replaceRoot - This final stage just recovers the original object by resetting the root to the `chart` field we added in the group operation.

Now we can add a stage before the first stage for the query using the `$match` pipeline, and we can add a final stage of `$count` or `$skip` and `$limit` to do the pagination. We may also be able to retrieve the count and paginated elements in the same query to optimise, probably using another `$addFields` stage, but I haven't checked for that.

Is it worth doing this over loading everything in memory, figuring out distinct charts and paginating client (server)-side? As you said, we're generally going to be dealing with a small enough collection that it probably won't make a huge difference. Also, there is some added work in supporting https://godoc.org/github.com/globalsign/mgo#Pipe in our datastore interface.

Personally, I think it is worth it to delegate the hard work to MongoDB (which is probably faster at doing this). It is less code, and with comments, I would argue simpler to understand what's going on. The filtering and uniquifying happens at the same time rather than separately, and so the result can clearly be understood by just looking at the single query.

That said, I think we need to find out exactly where the bottleneck is to understand if pagination will actually speed up the Hub. If the bottleneck is server -> database, then using this aggregated query will really help, because we avoid fetching more charts than we need to from the database. If loading all charts is actually pretty fast, and the bottleneck lies in sending this information back to the browser, then implementing pagination server-side should be fine (or it doesn't really matter whether this happens server-side or database). If sending all chart information to the browser also happens to be fast, and Angular/the browser is just slow at rendering all the cards and loading the icons, then we're better off just paginating browser-side in Angular and not touching the server-side at all. @andresmgot would you be able to spend some time investigating where the bottleneck is so we can decide on what the best solution is?

andresmgot: An update here:

 - Querying all the resources like we were doing is around 80 ms (in my local instance, minikube)
 - Querying just a page of 32 resources is around 20 ms
 - Counting all the documents (`db.charts.find(...).count()`) is only 2 ms so it's not relevant
 - Doing an aggregation of functions takes as you suggest takes 40 ms (still better than the first option)

I have checked that heaviest operations in the hub are (apart than the angular app loading) are the download of the app itself (~200 ms) and then the logo(s) fetching that can take up to 250 ms. In any case if we reduce a good chunk of the time taken in the chartsvc it will be noticeable. Also note that the data send in the request gets reduced from 170kb to 16kb.

I am still adapting the `datastore` library to support the `Pipe` method.

prydonius: Thanks for the research @andresmgot, this is very helpful! So, it sounds like we should go with the aggregation query to get an accurate count of unique charts whilst also paginating.

As I suspected, the major bottleneck is Angular loading all the charts, so reducing to 40ms sounds reasonable here as we're get a more noticeable increase when loading less charts in the frontend.

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,110 @@ type rel struct {
...
+
+// getPageNumberAndSize extracts the page number and size of a request. Default (1, 0) if not set
+func getPageNumberAndSize(req *http.Request) (int, int) {
+	page := req.FormValue("page")
+	size := req.FormValue("size")
+	pageInt, err := strconv.ParseUint(page, 10, 64)
+	if err != nil {
+		pageInt = 1
+	}
+	sizeInt, _ := strconv.ParseUint(size, 10, 64)
```
prydonius: add a comment above here that ParseUint will return 0 if size is not a positive integer

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,110 @@ type rel struct {
...
+	return int(pageInt), int(sizeInt)
+}
+
+func min(a, b int) int {
```
prydonius: Please add a comment above here as suggested in https://github.com/helm/monocular/pull/605/files#r261323150

cmd/chartsvc/handler.go
andresmgot: fair enough, extra information won't hurt

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,125 @@ type rel struct {
...
+// We are not using math.Min since that compares float64
+// and it's unnecesary complex.
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func uniqChartList(charts []*models.Chart) []*models.Chart {
```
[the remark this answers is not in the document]
andresmgot: We do, we are still using that in `listChartsWithFilters`. That is the function used to find a chart using a version and appversion. We don't need to do a pipeline there so we can use the simpler version which is a bit faster.

cmd/chartsvc/handler.go
```diff
@@ -61,34 +64,125 @@ type rel struct {
...
+		bson.M{"$replaceRoot": bson.M{"newRoot": "$chart"}},
+		// Order by name
+		bson.M{"$sort": bson.M{"name": 1}},
+	)
+
+	totalPages := 1
+	if pageSize != 0 {
+		// If a pageSize is given, returns only the the specified number of charts and
+		// the number of pages
+		length, err := countQuery.Count()
```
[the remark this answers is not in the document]
andresmgot: You are right, I missed that. The good news is that the repeated pipeline steps are cached so after executing "count" the retrieval of charts is faster.

To be fair, I have rechecked the times because I think this is getting a bit complex. The aggregation solution is slightly slower than querying everything, picking unique charts and do the pagination in the chartsvc (40 vs 50 ms). The good part of doing aggregation is that we need to maintain less code in our side, the downside is that we increase the vendor lock-in with MongoDB.

cmd/chartsvc/handler.go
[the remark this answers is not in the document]
prydonius: Ah okay, makes sense - thanks!

cmd/chartsvc/handler.go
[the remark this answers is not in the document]
prydonius: > To be fair, I have rechecked the times because I think this is getting a bit complex. The aggregation solution is slightly slower than querying everything, picking unique charts and do the pagination in the chartsvc (40 vs 50 ms). The good part of doing aggregation is that we need to maintain less code in our side (with the caveat that we need the code to filter unique charts anyway), the downside is that we increase the vendor lock-in with MongoDB.

This is a good analysis. One thing to consider is how this might scale if we had 1,000 charts, or 10, 000 - my assumption is MongoDB will be faster as it avoids having to send all those charts down the pipe to the chartsvc. There might also be some ways to improve the performance of the aggregation by configuring indexes in MongoDB.

For the same reasons as above, if we do end up switching to a different database in the future, we will likely want to reimplement the pipeline as a query for this new database (e.g. some sort of distinct SQL query if we go with a relational database). That said, this probably won't happen anytime soon.

Another solution we haven't considered yet is doing this filtering in `chart-repo`, though race conditions would make that more difficult (possibly solvable using transactions, though). Not saying we should consider this now, but worth keeping in mind.

cmd/chartsvc/handler.go
```diff
@@ -61,34 +63,127 @@ type rel struct {
...
 }
 
-// listCharts returns a list of charts
-func listCharts(w http.ResponseWriter, req *http.Request) {
+type meta struct {
+	TotalPages int `json:"totalPages"`
+}
+
+type count struct {
+	Count int `json:"count"`
```
prydonius: I believe this doesn't need the `json` tag since we never marshal it, right? Not hugely important, but I might accidentally think that we do marshal it based on this.

cmd/chartsvc/handler.go
```diff
@@ -61,34 +63,127 @@ type rel struct {
 	Links selfLink    `json:"links"`
 }
 
-// listCharts returns a list of charts
-func listCharts(w http.ResponseWriter, req *http.Request) {
+type meta struct {
+	TotalPages int `json:"totalPages"`
+}
+
+type count struct {
```
prydonius: maybe add a comment to say that this is used to retrieve a count from the database (might seem obvious, but it's not so obvious that we need a struct for this rather than just a variable)

andresmgot:
Thanks for the review @prydonius ! I still don't have write access to the repo so I cannot merge this.

PR #606, opened 2019-03-04
Removing emeritus maintainers from OWNERS

prydonius:
Removing @technosophos, @jackfrancis and @Angelmmiguel from OWNERS, as these maintainers have agreed to drop down to emeritus status.


jackfrancis:
lgtm ✌️

PR #609, opened 2019-03-07
Add search endpoint

andresmgot:
Requires #605

This PR adds a `/search` endpoint to the chartsvc to mimic the same behavior that monocular does in the client side:

https://github.com/helm/monocular/blob/master/frontend/src/app/shared/services/charts.service.ts#L75

I have tested the changes manually but since the function is mostly a mongodb query there is nothing valuable that we can add as unit tests.


cmd/chartsvc/handler.go
```diff
@@ -225,6 +225,43 @@ func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Para
...
+			{"repo.name": bson.M{"$regex": req.FormValue("match")}},
+			{"keywords": bson.M{"$elemMatch": bson.M{"$regex": req.FormValue("match")}}},
+			{"sources": bson.M{"$elemMatch": bson.M{"$regex": req.FormValue("match")}}},
+			{"maintainers": bson.M{"$elemMatch": bson.M{"name": bson.M{"$regex": req.FormValue("match")}}}},
+		},
+	}
+	if params["repo"] != "" {
+		conditions["repo.name"] = params["repo"]
+	}
+	if err := db.C(chartCollection).Find(conditions).All(&charts); err != nil {
```
prydonius: So if there are no matches MongoDB returns an error? Doesn't it just return an empty list with no error?

cmd/chartsvc/main.go
```diff
@@ -44,7 +44,9 @@ func setupRoutes() http.Handler {
 	apiv1 := r.PathPrefix(pathPrefix).Subrouter()
 	apiv1.Methods("GET").Path("/charts").Queries("name", "{chartName}", "version", "{version}", "appversion", "{appversion}").Handler(WithParams(listChartsWithFilters))
 	apiv1.Methods("GET").Path("/charts").HandlerFunc(listCharts)
+	apiv1.Methods("GET").Path("/charts/search").Queries("match", "{match}").Handler(WithParams(findCharts))
```
prydonius: can we call the query parameter `?q=`, that's more common than `match`

prydonius:
Looks like this PR is based on #605 as it's looking for `uniqChartList`?

`cmd/chartsvc/handler.go:261:29: undefined: uniqChartList`

andresmgot:
> Looks like this PR is based on #605 as it's looking for uniqChartList?

Yes, I have rebased master now that that PR is merged.

cmd/chartsvc/handler.go
```diff
@@ -321,6 +321,44 @@ func listChartsWithFilters(w http.ResponseWriter, req *http.Request, params Para
...
 }
 
+// findCharts returns the list of charts that matches the query param in any of these fields:
+//  - name
+//  - description
+//  - repository name
+//  - any keyword
+//  - any source
+//  - any maintainer name
+func findCharts(w http.ResponseWriter, req *http.Request, params Params) {
```
prydonius: one minor thing that I forgot to mention, it might make more sense to call this `searchCharts`?

andresmgot:
Merging this. Thanks for the review!

PR #615, opened 2019-03-13
Add truthiness check to arrayMatch

### LINKED ISSUE #614: Search is broken in some situations
The details are on https://github.com/helm/hub/issues/55

unguiculus:
Fixes #614 

Signed-off-by: Reinhard Nägele <unguiculus@gmail.com>

### DIFF
--- a/frontend/src/app/shared/services/charts.service.ts
+++ b/frontend/src/app/shared/services/charts.service.ts
@@ -87,7 +87,7 @@ export class ChartsService {
       return keywords.some((keyword) => {
         return !!keyword.match(re)
       })
-    }}
+    }
     
     return false
   }
--- a/frontend/src/app/shared/services/charts.service.ts
+++ b/frontend/src/app/shared/services/charts.service.ts
@@ -83,11 +83,13 @@ export class ChartsService {
   }
 
   arrayMatch(keywords: string[], re): boolean {
-    if(!keywords) return false
-
-    return keywords.some((keyword) => {
-      return !!keyword.match(re)
-    })
+    if (!!keywords) {
+      return keywords.some((keyword) => {
+        return !!keyword.match(re)
+      })
+    }}
+    
+    return false
   }
 
   /**


scottrigby:
From keybase chat: 

I'm thinking since we can't control the structure of charts that are in repos defined for a monocular instance, we should decide how we want to handle charts with missing data. 

Perhaps we should clearly define which chart data is required and which is optional. One way to address this could be to validate chart data, at which time charts missing required data could be removed from listing/searching, etc (`getCharts()`/`getChart()`). And charts missing optional data could be checked before trying to match each data point in `searchCharts`?

I'm not set up to test this right now, but since search is broken right now, if this fixes that issue perhaps we should merge now and discuss a larger fix later (whether the above, or some other proposal)?

unguiculus:
I agree with @scottrigby. This is a quick fix we should do anyways because I think a truthiness check makes sense here. But we need to collect requirements that must be met for charts to belisted.

scottrigby:
So after a quick look with someone here, it seems the issue is we call `match()` on members without checking first whether they're undefined. This is the error @prydonius saw in https://github.com/helm/hub/issues/55#issuecomment-471666795:
>TypeError: Cannot read property 'match' of undefined

scottrigby:
@unguiculus If that is correct ^ then perhaps we should do this instead? https://github.com/helm/monocular/pull/616 I need to run, so it was faster to just make a quick alternative PR. Can you take a look and see if this makes sense to you?

prydonius:
I might be missing something, but this PR doesn't seem to fix the issue. It looks like we're just changing the order in which we check for whether `keywords` is defined. The problem, as I understood it, is `keywords` might be a defined array, but some of its elements could be `undefined`. This means that the `keyword` inside the `keywords.some` block fails the `.match` call. This PR doesn't appear to fix that, but @scottrigby's #616 does.

Let me know if I'm misunderstanding the issue though, I haven't tested either PR.

prydonius:
cc @Dean-Coakley

unguiculus:
@prydonius I see, you're probably right. I was just kind of quickly trying the fix what you described in https://github.com/helm/hub/issues/55 and got it wrong.

PR #616, opened 2019-03-13
Check member is defined before calling match

### LINKED ISSUE #614: Search is broken in some situations

scottrigby:
Fixes #614

Alternative to #615 

For follow-up, this comment still applies https://github.com/helm/monocular/pull/615#issuecomment-472462431

Signed-off-by: Scott Rigby <scott@r6by.com>

### DIFF
--- a/frontend/src/app/shared/services/charts.service.ts
+++ b/frontend/src/app/shared/services/charts.service.ts
@@ -90,7 +90,7 @@ export class ChartsService {
     if(!keywords) return false
 
     return keywords.some((keyword) => {
-      return this.checkDefinedMatch(!!keyword, re)
+      return this.checkDefinedMatch(keyword, re)
     })
   }
 
--- a/frontend/src/app/shared/services/charts.service.ts
+++ b/frontend/src/app/shared/services/charts.service.ts
@@ -72,9 +72,9 @@ export class ChartsService {
     let re = new RegExp(query, 'i');
     return this.getCharts(repo).map(charts => {
       return charts.filter(chart => {
-        return chart.attributes.name.match(re) ||
-         chart.attributes.description.match(re) ||
-         chart.attributes.repo.name.match(re) ||
+        return this.checkDefinedMatch(chart.attributes.name, re) ||
+         this.checkDefinedMatch(chart.attributes.description, re) ||
+         this.checkDefinedMatch(chart.attributes.repo.name, re) ||
          this.arrayMatch(chart.attributes.keywords, re) ||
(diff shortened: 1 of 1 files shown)


prydonius:
cc @Dean-Coakley

prydonius:
@scottrigby do we know what chart repositories are missing some of the metadata that causes this to fail? Want to reproduce this with the right chart repository locally, to see if this fixes the issue.

scottrigby:
@prydonius I don't know which could be causing the issue 😑 However, based on the error in https://github.com/helm/hub/issues/55#issuecomment-471666795 (which I also see), and looking at the code, this appears to be the only place where `match` is called on an object member (the others appear to be directly on objects) 🤷‍♂️ 

To reproduce, what about a chart that only lists `email` (not `name`) in the `maintainers` map? I'm not set up to test, so that's just a guess.

Also, could `chart.attributes.repo.name.match(re)` be broken because of these two hub PRs that don't include a `name` in `repos.yaml`?
- https://github.com/helm/hub/pull/52
- https://github.com/helm/hub/pull/56

prydonius:
Thanks for the explanation @scottrigby, I agree that this should be a fairly safe change to make, so I'm happy to go ahead and merge this.

> Also, could chart.attributes.repo.name.match(re) be broken because of these two hub PRs that don't include a name in repos.yaml?

`repos.yaml` is purely metadata in the Hub repo, `repos-values.yaml` is what actually gets used in Monocular and this just links to the repo URL. The missing attributes are actually coming from the repo indexes which we don't easily have control over. https://github.com/helm/hub/pull/56 is still a worthy effort though to standardise that metadata.

frontend/src/app/shared/services/charts.service.ts
```diff
@@ -72,21 +72,25 @@ export class ChartsService {
...
+  checkDefinedMatch(member, matcher) {
+    return member && member.match(matcher);
+  }
+
   arrayMatch(keywords: string[], re): boolean {
     if(!keywords) return false
 
     return keywords.some((keyword) => {
-      return !!keyword.match(re)
+      return this.checkDefinedMatch(!!keyword, re)
```
prydonius: This was sending `true` to `checkDefinedMatch` and `true.match` is undefined. We actually want to send the string.

```suggestion
      return this.checkDefinedMatch(keyword, re)
```

scottrigby: Ah good call… fixed!

prydonius:
Thanks @scottrigby!!

PR #617, opened 2019-03-20
Add GoCenter as GOPROXY to speed up go builds

rimusz:
Signed-off-by: rimas <rmocius@gmail.com>

Using https://gocenter.io as GOPROXY will speed modules download and will ensure that modules referenced in `go.mod` are always the same, doesn't matter if some modules get deleted from GitHub and etc, they got cached for live at gocenter.io :-)


rimusz:
cc @migmartri @prydonius

go.mod
```diff
@@ -0,0 +1,44 @@
+module github.com/rimusz/monocular
```
prydonius: guess this should be github.com/helm/monocular?

rimusz: yes, autogeneration, will fix it up

rimusz: done

prydonius:
This is really great, thanks @rimusz. However, would you also be able to update the developer documentation to replace the dep references with the relevant go module commands? https://github.com/helm/monocular/blob/master/docs/development.md#developers-guide

rimusz:
sure

PR #619, opened 2019-03-30
Fix chart icon aspect ratio

Dean-Coakley:
## Issue
In the case that a chart icon is taller than it is wide it is forced to be square, ruining its aspect ratio.

## Examples:

image
From: https://hub.helm.sh/charts/stable/satisfy

image
From: https://hub.helm.sh/charts/nginx/nginx-ingress

## Fix:
image

image

### DIFF
--- a/frontend/src/app/chart-details/chart-details.component.scss
+++ b/frontend/src/app/chart-details/chart-details.component.scss
@@ -66,7 +66,7 @@ $icon-height: 150px;
       justify-content: center;
 
       img {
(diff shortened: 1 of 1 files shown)


Dean-Coakley:
@prydonius Can you take a look at this please?

prydonius:
@Dean-Coakley sorry for the delay in looking at this, I've been travelling and just got back from some time off! Thanks for spotting this issue and fixing it!!

Issue #566, opened 2018-11-17
CronJob Name Too Long - must be no more than 52 characters.


medined:
I ran a helm command to install monoclar:

```
./helm install --namespace office  monocular/monocular
```

which resulted in the following message.

```
Error: release wrapping-grasshopper failed: CronJob.batch "wrapping-grasshopper-monocular-sync-scheduled-incubator" is invalid: metadata.name: Invalid value: "wrapping-grasshopper-monocular-sync-scheduled-incubator": must be no more than 52 characters
```

mattiaperi:
Hi all, as @medined, I'm facing the same issue:
`$ helm install monocular/monocular
Error: release terrifying-grasshopper failed: CronJob.batch "terrifying-grasshopper-monocular-sync-scheduled-incubator" is invalid: metadata.name: Invalid value: "terrifying-grasshopper-monocular-sync-scheduled-incubator": must be no more than 52 characters`

`$ echo "terrifying-grasshopper-monocular-sync-scheduled-incubator" | wc -m
      58`

Obviously a:

`$ helm install --name monocular monocular/monocular`

forcing the release name less than 14 characters it's a workaround for the issue.

Dean-Coakley:
@mattiaperi I am working on resolving this issue, however I'm getting.
`must be no more than 63 characters`
and have validated that it works for me up to 63 characters. I installed a release with name of 18 chars.

Are both of you possibly on an old helm release?
```
$ helm version
Client: &version.Version{SemVer:"v2.13.0", GitCommit:"79d07943b03aea2b76c12644b4b54733bc5958d6", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.13.0", GitCommit:"79d07943b03aea2b76c12644b4b54733bc5958d6", GitTreeState:"clean"}
```

Dean-Coakley:
Nevermind, if the name is far too long it will fail on `spec.template.labels`
Otherwise it will fail on 
`metadata.name` of the cronjob at 52 chars as you've described

harindersingh:
If you're using helm 3,
`--name`
flag is invalid and will give
`Error: unknown flag: --name`

You directly provide name before chartname
`Usage:
  helm install [NAME] [CHART] [flags]`
`helm install monocular-release monocular/monocular`

Dean-Coakley:
Right, I'll try update the chart docs later today if I get a chance.

However that's unrelated to this issue:
`./helm install this-is-very-long-release-name chart/monocular`
[...]

PR #621, opened 2019-04-23
Fix CronJob name too long

Dean-Coakley:
fixes: https://github.com/helm/monocular/issues/566

# Changes:
* Cronjob name is now truncated at 52 characters
* Job name is now truncated at 63 characters

### DIFF
--- a/chart/monocular/templates/repo-sync-cronjobs.yaml
+++ b/chart/monocular/templates/repo-sync-cronjobs.yaml
@@ -2,7 +2,7 @@
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
-  name: {{ printf "%s-%s-%s" (include "fullname" $) "sync-scheduled" .name | trunc 52 }}
+  name: {{ printf "%s-%s" (include "fullname" $) "sync-scheduled" | trunc 42 }}-{{ .name | trunc 10 }}
   labels:
     monocular.helm.sh/repo-name: {{ .name }}
     app: {{ template "fullname" $ }}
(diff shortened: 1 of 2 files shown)


prydonius:
Thanks @Dean-Coakley! One issue I can see with this is if the chart release name is > 38 characters long all the CronJobs will have the same name, because the name of the repo is last. Though not likely, we can mitigate this by truncating separately to ensure at least part of the repo name is always there, e.g.:

```
name: {{ printf "%s-%s" (include "fullname" $) "sync-scheduled" | trunc 42 }}-{{ .name | trunc 10 }}
```

Of course, this also means that a repo named `tencharstr` and `tencharstr2` would also clash as they would both be truncated to `tencharstr`, but the likelihood of having such a large release name is also lower. We could also reserve a bigger portion for the repo name.

Dean-Coakley:
That makes a lot of sense

I've added truncation to job names too as it was necessary.
It now works for release names up to 44 characters. (Instead of the intended 53 max)

If you grep around for `template "fullname"` you'll see many, many cases of appending after 63 chars. Do you want me to solve these cases? I think this would be necessary to get up to to intended max of 53 chars.

prydonius:
@Dean-Coakley is the 53 max limit just on CronJob or on all resources? If all resources, then I agree that we should fix these other cases.

Dean-Coakley:
Could never find docs/code to describe at what char limit each resource should be truncated.

Closing for now, would be delighted to resume work with some guidance.

Issue #591, opened 2019-01-15
Charts with svg icons are not shown


andresmgot:
If a chart contains a SVG icon (e.g. `auditbeat` which icon is `https://www.elastic.co/assets/blt27d1fd26b0862613/icon-auditbeat-bb.svg`) it fails to get imported:

```
time="2019-01-15T12:13:43Z" level=error msg="failed to import icon" error="image: unknown format" name=auditbeat
```

This causes that the chart logo not to be shown.

cc/ @prydonius

Issue #577, opened 2018-12-12
SVG icons are not displayed


lambertjosh:
First, thanks for this great project! In doing a bit of testing, our `gitlab` chart uses an SVG for its icon. Unfortunately it is not displayed: https://hub.helm.sh/charts/gitlab/gitlab

According to the spec, SVG's are supported: https://github.com/helm/helm/blob/master/docs/charts.md#the-chartyaml-file

Would be great if SVG's were shown!

prydonius:
Thanks for the bug, this would be really cool! We currently download the icons and post-process them (currently just converting to PNG and resizing to 160x160), so we would need to change this to either convert the SVG to PNG so everything else stays the same, or store the SVGs and display them differently in the frontend. I'm not sure we really need to preserve the SVG, so converting to PNG might be the simplest option, but I'm not sure if there are any side effects from doing that conversion.

PR #631, opened 2019-05-23
Add support for SVG icons

andresmgot:

Fixes https://github.com/kubeapps/kubeapps/issues/1007
Fixes https://github.com/helm/monocular/issues/577
Fixes https://github.com/helm/monocular/issues/591

Add support to display to store icons in the database as SVG icons. We don't do any processing in that case, the frontend app is in charge of showing the image with a proper size.


cmd/chartsvc/handler.go
```diff
@@ -407,7 +413,7 @@ func chartVersionAttributes(cid string, cv models.ChartVersion) models.ChartVers
 
 func chartAttributes(c models.Chart) models.Chart {
 	if c.RawIcon != nil {
-		c.Icon = pathPrefix + "/assets/" + c.ID + "/logo-160x160-fit.png"
+		c.Icon = pathPrefix + "/assets/" + c.ID + "/logo"
```
andresmgot: I have changed the URL since the icon doesn't have to be a PNG anymore (nor be 160x160)

cmd/chartsvc/main.go
```diff
@@ -54,7 +54,7 @@ func setupRoutes() http.Handler {
 	apiv1.Methods("GET").Path("/charts/{repo}/{chartName}").Handler(WithParams(getChart))
 	apiv1.Methods("GET").Path("/charts/{repo}/{chartName}/versions").Handler(WithParams(listChartVersions))
 	apiv1.Methods("GET").Path("/charts/{repo}/{chartName}/versions/{version}").Handler(WithParams(getChartVersion))
-	apiv1.Methods("GET").Path("/assets/{repo}/{chartName}/logo-160x160-fit.png").Handler(WithParams(getChartIcon))
```
andresmgot: I was going to keep the previous path for backward compatibility but as far as I understand it's not needed because the web clients check the information of the chart and get the icon URL from it.

cmd/chart-repo/utils.go
```diff
@@ -310,20 +310,35 @@ func fetchAndImportIcon(dbSession datastore.Session, c chart) error {
...
+		b, err = ioutil.ReadAll(res.Body)
+		if err != nil {
+			return err
+		}
+		contentType = res.Header.Get("Content-Type")
+	} else {
+		// if the icon is in any other format try to convert it to PNG
+		orig, err := imaging.Decode(res.Body)
+		if err != nil {
+			log.Infof("failing to decode %v", err)
```
prydonius: This should be an `Errorf`, also can we add fields that describe what chart this icon is for? E.g. https://github.com/helm/monocular/pull/631/files#diff-dd464ab7a9989bed9c6a3801d51ade13R288

```suggestion
			log.WithFields(log.Fields{"name": c.Name}).WithError(err).Error("failed to decode icon")
```

cmd/chartsvc/main.go
prydonius: Perhaps we should just keep it around anyway, there might be cases where a client is not forming the URL from the API?

andresmgot: okay, added just in case

PR #637, opened 2019-07-10
[chart-repo] Store repositories checksum in cache to avoid reprocessing

andresmgot:

The `chart-repo` job, download, parses and populates the database even if there are no changes since the last execution. To be able to execute this job as often as needed we can store the checksum of a repository in the database and only process a repository when there are actual changes.

See: https://github.com/kubeapps/kubeapps/issues/1070

### DIFF
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -106,7 +106,7 @@ func syncRepo(dbSession datastore.Session, repoName, repoURL string, authorizati
 
 	// Check if the repo has been already processed
 	if repoAlreadyProcessed(dbSession, repoName, repoChecksum) {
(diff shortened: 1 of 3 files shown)


cmd/chart-repo/utils.go
```diff
@@ -85,14 +87,30 @@ func init() {
...
+	}
+
+	repoChecksum, err := getSha256(repoBytes)
+	if err != nil {
+		return err
+	}
+
+	// Check if the repo has been already processed
+	if repoAlreadyProcessed(dbSession, repoName, repoChecksum) {
+		log.WithFields(log.Fields{"url": repoURL}).Info("Skipping repository since it has been already processed")
```
prydonius: perhaps this should say since there are no new updates instead, processed implies a one-time thing

PR #640, opened 2019-07-19
add helm hub china mirror in readme

wonderflow:
Chinese user may have some problems to use Helm Hub, so we mirror it and replace unusable url to help developers to use Helm in China.

\cc @resouer

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,10 @@ Broker](https://blog.openshift.com/automation-broker-discovering-helm-charts/)).
 includes in-cluster features and can still be installed and used until your team
 has migrated to another tool.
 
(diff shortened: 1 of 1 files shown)


prydonius:
Thanks @wonderflow, I think it makes more sense to add this to the README in github.com/helm/hub, as this is just the source we use for the Helm Hub and AFAIK https://developer.aliyun.com/hub is not using this source.

mattfarina:
@wonderflow can you please make this change to the github.com/helm/hub repo instead.

wonderflow:
@mattfarina  @prydonius  OK，thanks

PR #643, opened 2019-09-26
Change for helm v3 client

mattfarina:
This PR contains two changes to support helm v3.

First, `helm search` will let one search the Helm Hub. But, the response has URLs which aren't installable. So, we want to make the monocular URLs installable.

This change makes it so that one can run a command like...

```
$ helm install https://monocular.example.com/charts/foo/bar
```

Monocular will redirect to the URL with the tgz file. Handles calls for provenance files as well

Second, Helm search is calling to the url at `/api/chartsvc/v1`. This exposes the internal architecture of Monocular. We'd prefer to have a public URL. So, this exposes the search at `/api/v1` as well. We want to keep both so that we could someday treat public calls differently from the UI calls if needed.


mattfarina:
Note, not sure of the right way to bump the version as this needs to go into a new monocular release for the full package to work.

andresmgot:
> Note, not sure of the right way to bump the version as this needs to go into a new monocular release for the full package to work.

It's a new feature, in theory the minor version should be bumped.

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
 
 	return cvl
 }
+
+func handleRedirect(w http.ResponseWriter, req *http.Request) {
+
```
prydonius: nit: empty line

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
 
 	return cvl
 }
+
+func handleRedirect(w http.ResponseWriter, req *http.Request) {
```
prydonius: perhaps this should be named `getChartVersionPackage`? Looks like most of the handlers in this file start with `get` or `list`. Another option is `redirectToChartVersionPackage`.

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+
+	// The URL can be in one of two forms:
+	// - /v1/redirect/charts/stable/aerospike
+	// - /v1/redirect/charts/stable/aerospike/v1.2.3
```
prydonius: I would split the handlers up and extract the shared logic to other functions. We should try to use the router as much as possible to handle path parameters.

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+	// And either of these can optionally have a trailing /
+
+	// Make sure the path is valid
+	ct := strings.TrimPrefix(req.URL.Path, "/v1/redirect/charts/")
+
+	// check if URL for provenance
+	prov := strings.HasSuffix(ct, ".prov")
```
prydonius: I think we may also want to consider handling the provenance file in a separate handler

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+
+	// Look it up. This will be different if there is a version or we are getting
+	// the latest
+	db, closer := dbSession.DB()
+	defer closer()
+	var chart models.Chart
+	chartID := fmt.Sprintf("%s/%s", p[0], p[1])
+
+	if version == "" {
+		if err := db.C(chartCollection).FindId(chartID).One(&chart); err != nil {
```
prydonius: Since we're starting to repeat these MongoDB calls, we may want to extract them out to separate functions and call them from here and in `getChart` (and the one below in `getChartVersion`)

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+			return
+		}
+	}
+
+	// Respond with proper redirect for tarball and prov
+	if len(chart.ChartVersions) > 0 {
+		cv := chart.ChartVersions[0]
+		if len(cv.URLs) > 0 {
+			if prov {
+				http.Redirect(w, req, cv.URLs[0]+".prov", http.StatusTemporaryRedirect)
```
prydonius: Just a thought, instead of redirecting, it might make sense to stream the file directly back to the client. I know the intention of this is for public repositories, but this could be useful for instances of Monocular pointing to private repositories that might not be accessible by the client.

cmd/chartsvc/handler.go
mattfarina: I used `redirectToChartVersionPackage` because it's not getting the version package... it's more getting a redirect to the package.

cmd/chartsvc/handler.go
mattfarina: nit fixed

cmd/chartsvc/handler.go
mattfarina: I'm not sure how to handle that with mux (see `setupRoutes`). `PathPrefix` is used because the redirect URL is going to have multiple path parts to it. In the router there is not method for filtering based on suffix, that I'm aware of. Do you have a suggestion?

cmd/chartsvc/handler.go
mattfarina: I thought about that but if we do that for the Helm Hub it means we are eating a lot more egress cost as we stream all the packages out from the hub. It also means more system resources are required. I was attempting to be cost sensitive with this implementation.

cmd/chartsvc/handler.go
mattfarina: So... here is my problem and maybe you can help with it. Let's say I have the path `my-repo/my-chart/0.1.0/`. Now we also need to get the provenance file for that. This would be `my-repo/my-chart/0.1.0/.prov`. Notice the placements of the path separators. Or, what if we have the path `my-repo/my-chart/0.1.0` and need to get `my-repo/my-chart/0.1.0.prov`. How do we separate the chart tarball from the provenance one with mux?

There is some variation to the possible inputs which makes handling some of this in the router difficult. Any pointers on that?

cmd/chartsvc/handler.go
```diff
@@ -444,3 +445,75 @@ func newChartVersionListResponse(c *models.Chart) apiListResponse {
...
+
+	// check if URL for provenance
+	prov := strings.HasSuffix(ct, ".prov")
+	ct = strings.TrimSuffix(ct, ".prov")
+
+	ct = strings.TrimSuffix(ct, "/") // Removing the optional / on the end
+	p := strings.Split(ct, "/")
+
+	// Not enough parts passed in to the path
+	if len(p) < 2 || len(p) > 3 {
```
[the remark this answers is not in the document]
mattfarina: Updated to make it more clear

cmd/chartsvc/handler.go
mattfarina: Can we do this as part of a separate change? This query is used 4 times by 4 different callbacks. If the logic were to be extracted I would suggest doing it uniformly and updating all of them. It's worth being tested separately from this feature addition IMHO. Repeated the same database query is common here as it stands.

Issue #636, opened 2019-07-02
Update chart to stable resources


mattfarina:
The resources in the chart are pre-GA (e.g., `extensions/v1beta1`). GA versions exist (e.g., `apps/v1`) and Kubernetes is planning to stop accepting the older versions at the API in the near future. So, the chart should be updated to the new resource versions.

PR #644, opened 2019-09-28
Update chart resources apiVersion

Dean-Coakley:
Fixes: https://github.com/helm/monocular/issues/636

### Issue:
This is a required change to make the chart installable on Kubernetes 1.16.

### Proposed changes:
* Update all `Deployment` resources of the chart to `apps/v1` as required by Kubernetes 1.16
* Update `mongodb` chart dependency. From `4.6.2` -> `7.2.10`.
  * This was required because `mongodb:4.6.2` also used the deprecated `"extensions/v1beta1"` apiVersion.
* Add `spec.selector` as required by the new apiVersion.
  
  > Error: release monocular-dev failed: Deployment.apps "monocular-dev-monocular-ui" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{"app":"monocular-dev-monocular-ui"}: `selector` does not match template `labels`]

#### Note: 
I did not increment the `version` of the chart. It looks like this is done via CI?

### DIFF
--- a/chart/monocular/Chart.yaml
+++ b/chart/monocular/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: monocular
 description: Monocular is a search and discovery front end for Helm Charts Repositories.
-version: 1.4.10
+version: 1.4.12
 appVersion: v1.8.3
 home: https://github.com/helm/monocular
 sources:
--- a/chart/monocular/requirements.lock
+++ b/chart/monocular/requirements.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: mongodb
   repository: https://kubernetes-charts.storage.googleapis.com
-  version: 4.6.2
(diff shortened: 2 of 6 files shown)


Dean-Coakley:
@mattfarina @absoludity Are either of ye able to review/approve/merge?

prydonius:
The CI only bumps the chart version if we have tagged a release for new code changes, but chart changes that aren't tied to a code release should manually bump the version. I think this is a non-breaking change, so we should bump the patch version of the chart here.

Dean-Coakley:
@prydonius Sure. I wasn't certain if updating mongodb dep version was a breaking change.

Thanks for the review. I've bumped the patch version of the chart.

Dean-Coakley:
Rebumped chart version 2 patches - under the assumption that #647 will be merged first

PR #648, opened 2019-10-04
[Proposal] Increase how often charts are synced by default

Dean-Coakley:
Syncing charts every hour seems pretty low.

#### Proposed changes
* Decrease delay between syncs from `1 hour`->`10 minutes`

I noticed that with kubeapps/kubeapps the default is every 10 minutes.
https://github.com/kubeapps/kubeapps/blob/0df09514088ed41592e9e613a79dc9b432cea751/chart/kubeapps/values.yaml#L104-L105


Dean-Coakley:
Bump patch version assuming that https://github.com/helm/monocular/pull/644 will be merged before this.

prydonius:
cc @mattfarina - would this be okay for the Helm Hub?

Dean-Coakley:
@mattfarina Can you take a look at this when you get a chance?

PR #655, opened 2019-11-26
Delete repository from all collections

andresmgot:
Fixes: https://github.com/kubeapps/kubeapps/issues/1161

When deleting the repo, we should also delete the document from the `repos` collection.


andresmgot:
_merging this as trivial_

Issue #657, opened 2020-02-11
Have chart-repo filter charts when syncing


gregsidelinger:
Need a way to limit what charts are displayed when syncing with a repository that contains charts users should use.

I played around with a --filter-name option to only include charts that match the names passed in.  I also played around a --filter-annotation flag to only sync repos that contained an annotation of a specific value.

PR #658, opened 2020-02-11
Adding filter options to chart-repo

gregsidelinger:
Signed-off-by: Greg Sidelinger <gate@ilive4code.net>

Here are my basic filter changes mentioned in https://github.com/helm/monocular/issues/657

### DIFF
--- a/cmd/chart-repo/chart_repo.go
+++ b/cmd/chart-repo/chart_repo.go
@@ -47,7 +47,7 @@ func init() {
 		cmd.Flags().String("mongo-url", "localhost", "MongoDB URL (see https://godoc.org/github.com/globalsign/mgo#Dial for format)")
 		cmd.Flags().String("mongo-database", "charts", "MongoDB database")
 		cmd.Flags().String("mongo-user", "", "MongoDB user")
-		cmd.Flags().StringSliceVar(&filterAnnotations, "filter-annotation", []string{}, "Filter charts with annotation")
+		cmd.Flags().StringSliceVar(&filterAnnotations, "filter-annotation", []string{}, "Filter by charts that match any of these annotations")
 		cmd.Flags().StringSliceVar(&filterNames, "filter-name", []string{}, "Filter charts name")
 
 		// see version.go
--- a/cmd/chart-repo/chart_repo.go
+++ b/cmd/chart-repo/chart_repo.go
@@ -39,12 +39,17 @@ func main() {
 
 func init() {
 	cmds := []*cobra.Command{syncCmd, deleteCmd}
+	filterAnnotations := []string{}
+	filterNames := []string{}
 
 	for _, cmd := range cmds {
 		rootCmd.AddCommand(cmd)
 		cmd.Flags().String("mongo-url", "localhost", "MongoDB URL (see https://godoc.org/github.com/globalsign/mgo#Dial for format)")
 		cmd.Flags().String("mongo-database", "charts", "MongoDB database")
 		cmd.Flags().String("mongo-user", "", "MongoDB user")
+		cmd.Flags().StringSliceVar(&filterAnnotations, "filter-annotation", []string{}, "Filter charts with annotation")
+		cmd.Flags().StringSliceVar(&filterNames, "filter-name", []string{}, "Filter charts name")
+
 		// see version.go
 		cmd.Flags().StringVarP(&userAgentComment, "user-agent-comment", "", "", "UserAgent comment used during outbound requests")
 		cmd.Flags().Bool("debug", false, "verbose logging")
--- a/cmd/chart-repo/chart_repo.go
+++ b/cmd/chart-repo/chart_repo.go
@@ -48,7 +48,7 @@ func init() {
 		cmd.Flags().String("mongo-database", "charts", "MongoDB database")
 		cmd.Flags().String("mongo-user", "", "MongoDB user")
 		cmd.Flags().StringSliceVar(&filterAnnotations, "filter-annotation", []string{}, "Filter by charts that match any of these annotations")
-		cmd.Flags().StringSliceVar(&filterNames, "filter-name", []string{}, "Filter charts name")
+		cmd.Flags().StringSliceVar(&filterNames, "filter-name", []string{}, "Filter by charts that match these names")
 
 		// see version.go
 		cmd.Flags().StringVarP(&userAgentComment, "user-agent-comment", "", "", "UserAgent comment used during outbound requests")
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -31,6 +31,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -562,7 +563,8 @@ func filterEntry(entry *helmrepo.ChartVersion, filter *filters) bool {
 
 	if len(filter.Names) > 0 {
 		for _, n := range filter.Names {
-			if entry.Name == n {
+			matched, _ := filepath.Match(n, entry.Name)
+			if matched {
 				return true
 			}
 		}
--- a/cmd/chart-repo/utils.go
+++ b/cmd/chart-repo/utils.go
@@ -281,11 +281,10 @@ func chartsFromIndex(index *helmrepo.IndexFile, r repo, filter *filters) []chart
 
 		if len(filter.Annotations) > 0 ||
 			len(filter.Names) > 0 {
-			if filterEntry(entry[0], filter){
-				charts = append(charts, newChart(entry, r))
+			if !filterEntry(entry[0], filter) {
+				log.WithFields(log.Fields{"name": entry[0].GetName()}).Info("skipping chart as filters did not match")
+				continue
 			}
-			log.WithFields(log.Fields{"name": entry[0].GetName()}).Info("skipping chart as filters did not match")
-			continue
 		}
 
 		charts = append(charts, newChart(entry, r))
@@ -548,7 +547,7 @@ func initNetClient(additionalCA string) (*http.Client, error) {
 }
 
 // return true if entry matches any filter
-func filterEntry( entry *helmrepo.ChartVersion, filter *filters) bool {
+func filterEntry(entry *helmrepo.ChartVersion, filter *filters) bool {
 	if len(filter.Annotations) > 0 {
 		for a, av := range filter.Annotations {
 			if v, ok := entry.Annotations[a]; ok {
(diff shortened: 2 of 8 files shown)


cmd/chart-repo/utils_test.go
```diff
@@ -285,16 +285,29 @@ func Test_parseRepoIndex(t *testing.T) {
 func Test_chartsFromIndex(t *testing.T) {
 	r := repo{Name: "test", URL: "http://testrepo.com"}
 	index, _ := parseRepoIndex([]byte(validRepoIndexYAML))
-	charts := chartsFromIndex(index, r)
+	charts := chartsFromIndex(index, r, new(filters))
 	assert.Equal(t, len(charts), 2, "number of charts")
+	// filter on name
```
prydonius: could we split this up into sub-tests using `t.Run`? (e.g. https://github.com/helm/monocular/blob/4043fe3fd98409933f8500559cec216e49af92a5/cmd/chartsvc/handler_test.go#L73, but without the for-loop)

cmd/chart-repo/utils_test.go
```diff
@@ -285,16 +285,29 @@ func Test_parseRepoIndex(t *testing.T) {
...
+	filter.Names = append(filter.Names, "not-found")
+    charts = chartsFromIndex(index, r, filter)
+    assert.Equal(t, len(charts), 1, "number of charts")
+	// filter on annotation
+	filter = new(filters)
+	filter.Annotations = make(map[string]string)
+	filter.Annotations["sync"] = "true"
+	filter.Annotations["not-found"] = "missing"
+	filter.Annotations["no-value"] = ""
+    assert.Equal(t, len(charts), 1, "number of charts")
```
prydonius: what chart does this test pick up? I can't see any annotations in https://github.com/helm/monocular/blob/master/cmd/chart-repo/testdata/valid-index.yaml?

absoludity:
> Thanks @gregsidelinger, I like the idea! cc @andresmgot @absoludity

Nice. Yes storing extra data for charts when they are synced, and being able to query based on that data is also something we're needing in Kubeapps (though we're now using a different codebase for sync which supports postgresql as a backend).

cmd/chart-repo/utils_test.go
gregsidelinger: My bad, I was doing some cleanup and the updated test yam l did got get committed.  It will be there in the next push.

cmd/chart-repo/utils_test.go
```diff
@@ -285,16 +285,29 @@ func Test_parseRepoIndex(t *testing.T) {
...
 	r := repo{Name: "test", URL: "http://testrepo.com"}
 	index, _ := parseRepoIndex([]byte(validRepoIndexYAML))
-	charts := chartsFromIndex(index, r)
+	charts := chartsFromIndex(index, r, new(filters))
 	assert.Equal(t, len(charts), 2, "number of charts")
+	// filter on name
+	filter := new(filters)
+	filter.Names = append(filter.Names, "wordpress")
+	filter.Names = append(filter.Names, "not-found")
+    charts = chartsFromIndex(index, r, filter)
```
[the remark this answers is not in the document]
gregsidelinger: done

cmd/chart-repo/utils_test.go
gregsidelinger: I broke them out into separate tests but did not use a loop using r.Run

cmd/chart-repo/utils.go
```diff
@@ -271,13 +271,43 @@ func parseRepoIndex(body []byte) (*helmrepo.IndexFile, error) {
...
 		if entry[0].GetDeprecated() {
 			log.WithFields(log.Fields{"name": entry[0].GetName()}).Info("skipping deprecated chart")
 			continue
 		}
+		if len(filter.Annotations) > 0 {
+			for a, av := range filter.Annotations {
+				if v, ok := entry[0].Annotations[a]; ok {
+					if len(av) == 0 {
+						charts = append(charts, newChart(entry, r))
+						continue ENTRYLOOP
```
prydonius: I think we can avoid this and improve the readability here by splitting out these filters into separate functions. Also, there might be a bug here. What if I have a name filter and an annotation filter, only the annotation filter will hold?

Something like:

```
if !matchesAnnotationsFilter(filter.Annotations, entry) || !matchesNamesFilter(filter.Names, entry) {
  // log skipping chart
  continue
}

charts = append(charts, newChart(entry, r))
```

Those functions can return true if filter.Annotations or filter.Names is empty.

WDYT?

cmd/chart-repo/utils_test.go
```diff
@@ -285,19 +285,62 @@ func Test_parseRepoIndex(t *testing.T) {
...
+func Test_chartsFromIndexFilterByAnnotationDuplicateMatches(t *testing.T) {
+	r := repo{Name: "test", URL: "http://testrepo.com"}
+	index, _ := parseRepoIndex([]byte(validRepoIndexYAML))
+	filter := new(filters)
+	filter.Annotations = make(map[string]string)
+	filter.Annotations["sync"] = "true"
+	filter.Annotations["sync-by-name-only"] = ""
+	charts := chartsFromIndex(index, r, filter)
+	assert.Equal(t, len(charts), 1, "number of charts")
+}
```
prydonius: let's add a test for both an annotation and name filter

cmd/chart-repo/utils_test.go
```diff
@@ -285,19 +285,62 @@ func Test_parseRepoIndex(t *testing.T) {
...
+
+func Test_chartsFromIndexFilterByAnnotationWithValue(t *testing.T) {
+	r := repo{Name: "test", URL: "http://testrepo.com"}
+	index, _ := parseRepoIndex([]byte(validRepoIndexYAML))
+	filter := new(filters)
+	filter.Annotations = make(map[string]string)
+	filter.Annotations["sync"] = "true"
+	filter.Annotations["not-found"] = "missing"
```
prydonius: Is the expectation here that it only needs to match one of the annotations and not all of the annotations? I would have assumed that all the annotations need to match for the entry to be selected (I believe this is also how `kubectl` label filters work).

cmd/chart-repo/utils_test.go
```diff
@@ -285,19 +285,62 @@ func Test_parseRepoIndex(t *testing.T) {
...
+	filter.Annotations["sync-by-name-only"] = ""
```
prydonius: Is this not testing the same thing as `Test_chartsFromIndexFilterByAnnotationDuplicateMatches`?

cmd/chart-repo/utils_test.go
gregsidelinger: Good point I did not think about that.  My intend was if any of them matched it synced.  That is actually how I just rewrote the code as a function.

```
// filterEntry return true if entry matches a filter attribute
func filterEntry( entry *helmrepo.ChartVersion, filter *filters) bool 
````

File: cmd/chart-repo/utils_test.go (hunk 7 above)
Comment (gregsidelinger) [reply cmd/chart-repo/utils_test.go]: Yes, `filter.Annotations["sync-by-name-only"] = ""` should not be there, too much copy and pasting.

File: cmd/chart-repo/utils_test.go (hunk 5 above)
Comment (gregsidelinger) [reply cmd/chart-repo/utils_test.go]: done

File: cmd/chart-repo/utils.go (hunk 4 above)
Comment (gregsidelinger) [reply cmd/chart-repo/utils.go]: Done, I was thinking the same thing when using the ENTRYLOOP label but was being lazy.  Since my intention was to sync if any of the filters matched I did a single function that returns true if a filter matched.

File: cmd/chart-repo/utils.go [hunk 8]
```diff
@@ -271,13 +271,23 @@ func parseRepoIndex(body []byte) (*helmrepo.IndexFile, error) {
...
 	for _, entry := range index.Entries {
 		if entry[0].GetDeprecated() {
 			log.WithFields(log.Fields{"name": entry[0].GetName()}).Info("skipping deprecated chart")
 			continue
 		}
+
+		if len(filter.Annotations) > 0 ||
+			len(filter.Names) > 0 {
+			if filterEntry(entry[0], filter){
+				charts = append(charts, newChart(entry, r))
```
prydonius: nitpick: it might be simpler to flip this condition and continue if it enters it, and just fallback to the append on line 291 if it matches the filter. i.e. similar to the deprecation condition above

cmd/chart-repo/utils.go
```diff
@@ -271,13 +271,23 @@ func parseRepoIndex(body []byte) (*helmrepo.IndexFile, error) {
...
+			}
+			log.WithFields(log.Fields{"name": entry[0].GetName()}).Info("skipping chart as filters did not match")
```
prydonius: This would need to be in an else block?

cmd/chart-repo/chart_repo.go
```diff
@@ -39,12 +39,17 @@ func main() {
...
 	cmds := []*cobra.Command{syncCmd, deleteCmd}
+	filterAnnotations := []string{}
+	filterNames := []string{}
 
 	for _, cmd := range cmds {
 		rootCmd.AddCommand(cmd)
 		cmd.Flags().String("mongo-url", "localhost", "MongoDB URL (see https://godoc.org/github.com/globalsign/mgo#Dial for format)")
 		cmd.Flags().String("mongo-database", "charts", "MongoDB database")
 		cmd.Flags().String("mongo-user", "", "MongoDB user")
+		cmd.Flags().StringSliceVar(&filterAnnotations, "filter-annotation", []string{}, "Filter charts with annotation")
```
prydonius: Given the intent to match any annotation (and not all), we can clarify that here:

```suggestion
		cmd.Flags().StringSliceVar(&filterAnnotations, "filter-annotation", []string{}, "Filter by charts that match any of these annotations")
```

cmd/chart-repo/chart_repo.go
```diff
@@ -39,12 +39,17 @@ func main() {
...
+		cmd.Flags().StringSliceVar(&filterNames, "filter-name", []string{}, "Filter charts name")
```
prydonius:
```suggestion
		cmd.Flags().StringSliceVar(&filterNames, "filter-name", []string{}, "Filter by charts that match these names")
```

cmd/chart-repo/utils.go
gregsidelinger: Done.  I ended up adding more test data to valid-index.yaml as I removed the ! from  `if !filterEntry(entry[0], filter)` to make the test fail and realized with only two charts of data the test would pass no mater what when filtering.

gregsidelinger:
gofmt looks happy on my end.

prydonius:
Thanks!

prydonius:
@andresmgot do you want to create a new Monocular release with this change?

andresmgot:
sure, thanks for the review!

PR #660, opened 2020-02-12
Update helm install to Helm 3

### LINKED ISSUE #659: Update to Helm 3 the 'helm install' instruction
The `helm install` command doesn't work in Helm 3.

Screen Shot 2020-02-12 at 14 17 03

Using this instruction:
```bash
▶ helm install bitnami/airflow --version 4.3.1
Error: must either provide a name or specify --generate-name
```
My proposal is to add the `--generate-name` flag:
```bash
▶ helm install bitnami/airflow --version 4.3.1 --generate-name
```

Then the other instructions depend on the `README.md` (not maintained by Helm itself), but it would be nice to, at least, update to Helm 3 the common sections.

carrodher:
Signed-off-by: Carlos Rodriguez Hernandez <crhernandez@bitnami.com>

Update the `helm install` instruction to work on Helm 3

Fixes #659

### DIFF
--- a/frontend/src/app/chart-details/chart-details-usage/chart-details-usage.component.ts
+++ b/frontend/src/app/chart-details/chart-details-usage/chart-details-usage.component.ts
@@ -48,6 +48,6 @@ export class ChartDetailsUsageComponent implements OnInit {
   }
 
   get installInstructions(): string {
-    return `helm install ${this.chart.id} --version ${this.currentVersion}`;
+    return `helm install ${this.chart.id} --version ${this.currentVersion} --generate-name`;
   }
 }


prydonius:
This won't work for Helm 2 though, right? Maybe using `--name` is better?

carrodher:
> This won't work for Helm 2 though, right? Maybe using `--name` is better?

AFAIK there is not a compatible command that works in both cases so `--name` will work with Helm 2 but not Helm 3.

Those are the options:
- Random name:
```
▶ helm2 install bitnami/fluentd
▶ helm3 install --generated-name bitnami/fluentd
```
- Defined name:
```
▶ helm2 install --name myChart bitnami/fluentd
▶ helm3 install myChart bitnami/fluentd
```

prydonius:
😬 

@mattfarina what do you suggest we do here?

andyfeller:
@carrodher @mattfarina : where does this stand?

